PR #6, opened 2021-06-16
Docker compose config updates

Windvis:
This
- doesn't restart the services while developing
- doesn't start the basic auth check service while developing
- adds a .env.example file


Windvis:
If you create a .env file with a `COMPOSE_FILE` value you can just type `docker-compose some-command` without having to list the compose files each time.

Windvis:
Can we merge this? Any blockers?

PR #7, opened 2021-06-24
Change in the auth config to enable edit

anitacaron:


config/authorization/config.ex
```diff
@@ -11,36 +11,47 @@ defmodule Acl.UserGroups.Config do
     [
       %GroupSpec{
         name: "public",
-        useage: [:read],
+        useage: [:read, :write, :read_for_write],
```
bdevloed, line 14: Not an expert on this. But doesn't this config, combined with the identifier param in docker-compose.yml allow unauthenticated users to write? There's currently a basic auth preventing this, but I assume this will change in the near future when acm/idm gets integrated?
I was expecting a different graph for edits, cf https://github.com/lblod/app-digitaal-loket/blob/master/config/authorization/config.ex (but perhaps better to check with an expert on the semantic.works stack)

anitacaron: Yes, this is only for test purpose. We have to change when we integrate with acm.

PR #14, opened 2021-07-07
Include inheritance between admin unit and worship/central services

anitacaron:


Windvis (approved):
Not my area of expertise but everything seems to work great with the frontend PR! Nice!

PR #15, opened 2021-07-09
Re-structure ministers roles resource

anitacaron:


config/dispatcher/dispatcher.ex
```diff
@@ -153,6 +161,30 @@ defmodule Dispatcher do
...
 
+  match "/minister-positions/*path", %{ accept: [:json], layer: :api} do
+    Proxy.forward conn, path, "http://resource/minister-positions/"
+  end
+
+  match "/minister-position-functions/*path", %{ accept: [:json], layer: :api} do
+    Proxy.forward conn, path, "http://resource/minister-position-functions/"
+  end
+
+  match "/financering-codes/*path", %{ accept: [:json], layer: :api} do
```
Windvis, line 172: This sounds like the dutch `financiering` but misspelled :smile:. I guess this should be `financing-code` instead? Must have missed that in the previous PR since the domain.json file already includes that. sorry :smile:.

anitacaron: That's good! Better late than never 😅

Windvis (approved):

config/dispatcher/dispatcher.ex
```diff
@@ -1,5 +1,7 @@
 defmodule Dispatcher do
   use Matcher
+organisaties-data Private
```
Windvis, line 3: WAAAAIT, how did that get in

Windvis: I'm not using the Github merge interface again :smile:

PR #16, opened 2021-07-09
Add site type to site resource

anitacaron:


config/resources/domain.json
```diff
@@ -846,6 +851,16 @@
...
+    },
+    "site-types": {
+      "class": "code:TypeVestiging",
+      "attributes": {
+        "label": {
+          "type": "string",
+          "predicate": "skos:prefLabel"
+        }
+      },
+      "new-resource-base": "http://lblod.data.gift/concepts/"
```
Windvis, line 863: Just double checking. Is this correct? Seems to be the same as on line 853.

anitacaron: Do you mean the "new-resource-base"? They're all the same for the code list, not going to create via the application. Waiting for Niels to check.

Windvis: Ok! Just making sure since it's the first time I noticed it :smile:.

Windvis (approved):

PR #17, opened 2021-07-12
kalliope api version

bdevloed:
This release includes a fix to handle the same triple in the insert and delete part of a single delta msgs and a different bulk upload mechanism. cf docs


bdevloed:
Ok to deploy on dev?

PR #29, opened 2021-08-13
Unique list of resource types

nbittich:


config/authorization/config.ex
```diff
@@ -70,6 +33,7 @@ defmodule Acl.UserGroups.Config do
                         "http://schema.org/ContactPoint",                          
                         "http://data.lblod.info/vocabularies/leidinggevenden/Bestuursfunctie",
                         "http://data.lblod.info/vocabularies/erediensten/CentraalBestuurVanDeEredienst",
+                        "http://data.lblod.info/vocabularies/erediensten/CentraleBestuursorgaan",
```
anitacaron, line 36: We're not using this type. Why is it back?

nbittich: in your previous change you let it in one side and removed it in the other :P I saw the diff so I took the most complete array. I'll fix it

anitacaron: yes, too many places to put the same information.

nbittich: I removed it, now there is just one place to put the resource type to avoid the confusion in the future

config/authorization/config.ex
```diff
@@ -7,45 +7,8 @@ alias Acl.GroupSpec, as: GroupSpec
...
-                  } LIMIT 1",
-          vars: []
-        },
-        graphs: [
-          %GraphSpec{
-            graph: "http://mu.semte.ch/graphs/contacthub/141d9d6b-54af-4d17-b313-8d1c30bc3f5b/ChAdmin",
-            constraint: %ResourceConstraint{
-              resource_types: [
-                        "http://www.w3.org/ns/org#Organization",
+ @protectected_resource_type [
```
anitacaron, line 10: Do you mean `protected`?

anitacaron (changes requested):

anitacaron (approved):

PR #30, opened 2021-08-20
docker-compose config refactoring

Windvis:
This removes the prod config file so that the project is in line with the other lblod apps.

There are some other changes that can be rolled back or split into a separate pr if needed:
- remove the metis services since they aren't used anymore.
- ~disable certain services during local development~ rolled back since it requires a recent docker-compose version.


Windvis:
It seems the profiles functionality that's used to disable services during development requires a fairly new docker-compose version which might be a problem for people developing on LTS distros.

madnificent:
This looks sensible to me.  Diving into this, I noticed a few other strange bits which we should tackle too and I have created #14 for those.  My changes involve whitespace.

I did not try to run this PR yet but foresee no issues.

nbittich:
Can you checkout this branch on dev and see if everything still works fine? if it's the case I'll approve it, I think the changes there are too big and we need to make sure once we set it on QA/prod, it still will works fine

bdevloed:
Getting 502 on calls to the jsonld API. Will have a look at it

bdevloed:
502 was due to the nginx server_name config being overwritten. i.e. not an issue introduced by the PR.

bdevloed (approved):

Windvis:
Can this be merged or is there something that needs to be tested more? Since I only copy pasted things around in the production configuration I don't think it will cause errors. And I've been using the dev version without issues.

nbittich:
Ok for me, can you update QA / dev once it's merged? thx

madnificent:
Yay for cleanup.  Thanks :)

PR #47, opened 2021-09-21
add privacy centric service

nbittich:


nbittich:
20210729182257-reason-code.ttl : this migration is temporary because we don't have the codelist yet I think

bdevloed (changes requested):
I would change URI's to `http://data.lblod.info/id/information-request-reason/{some-uuid}`

bdevloed (approved):

PR #54, opened 2021-10-10
Add x-frame-options when hosting frontend pages

madnificent:
We don't want these pages to be contained in a frame.  The x-frame-options serves that use.

To my understanding, it's fine if this is executed for all html pages.  Any attack would need to confuse people, and hence would need them to be visit a page.  Hence this is the right place to my understanding.  Feel free to verify that assumption.


Windvis:
I added the `always` param and made sure the header was still sent properly. :+1:

PR #60, opened 2021-11-03
changes for crab

nbittich:
Not really sure if that's how you would do it, @bdevloed can you have a look before I start the frontend part?

Thx!


bdevloed:
Looks good to me.
Other skos concepts in the project have an additional class, which is used in the domain.json. (cf all the code: entries in domain.json and their respective migration files).
However, I don't see harm in using a generic concept entity.

bdevloed (approved):

nbittich:
> Looks good to me. Other skos concepts in the project have an additional class, which is used in the domain.json. (cf all the code: entries in domain.json and their respective migration files). However, I don't see harm in using a generic concept entity.

Thanks, while implementing the frontend I found some issues that I fixed. Could you check again to make sure I didn't forget anything?

config/migrations/2021110315263111111111111111-source-crab.ttl
```diff
@@ -0,0 +1,13 @@
+INSERT DATA {
```
Windvis, line 1: I think this needs to be a `.sparql` file? :smile:

Windvis (approved):

PR #62, opened 2021-11-09
Feature/yasgui

nbittich:


bdevloed (changes requested):
I would just run this on another port in the dev config and not add this to the routing of the dispatcher. Hence it's easily switched off and on.

This let's a user bypass other auth checks and execute sparql queries directly.

Let me know if you see it differently.

nbittich:
> I would just run this on another port in the dev config and not add this to the routing of the dispatcher. Hence it's easily switched off and on.
> 
> This let's a user bypass other auth checks and execute sparql queries directly.
> 
> Let me know if you see it differently.

It has to be for all users and environment i think => https://binnenland.atlassian.net/browse/OP-846

bdevloed:
@nbittich : Thanks for the clarification. cfr comment on Jira.

I will wait for the results of that discussion.

bdevloed (approved):
Works as expected now.
i.e. SSN numbers do etc not leak through /sparql.

PR #65, opened 2021-11-15
forbid privacy centric data

nbittich:
not tested yet but this should prevent frontend to be able to save privacy centric data direcly, and also not sending them back anymore without using the privacy centric service.


bdevloed (commented):
Waiting for functional feedback on breaking the create new person functionality.

Windvis:
Should we just remove those existing fields? They are going to be replaced by the new flows anyway?

PR #66, opened 2021-11-16
Added two tweaked database settings

cecemel:
Note: virtuoso-production.ini might currently be overkill. Could
consider not using it.


bdevloed (approved):
Thx!

PR #67, opened 2021-11-16
Feature/delta producer administrative units

cecemel:
This PR is an example of how the delta-producers could be injected in the application. 
The focus would be here administrative-units, which could be extremely useful in many of the applications within the loket-realm.
The win would be is that no specific consumers need to be written in the loket realm to ingest the data.

I tried to explain per commit what does what. 

Extra notes:
- To keep it simple, I didn't introduce the reporting service, which I consider useful, as this gives us alerts when things go wrong. So I would recommend to use it too, if this goes to prod.
-  There are multiple background jobs, which follow the jobs model. I didn't integrate the jobs-dashboard, but it might be useful to have this one too, as this provides a GUI on the state of these multiple jobs.
-  The configuration is on what to export is very basic, and may be extended. There are many configuration options, and direct filters possible.
- The main assumption here; the data shared is public. So the endpoints which provide delta json files and dumps are accessible by everyone. There is currently work being done in semantic.works to be able to shield this in a structural way
- Nevertheless on first sight, a lot of the components are compatible with jsonld-delta-service and its security constraints. The extra services introduced here, could be introduced too for the data to share with kalliope. The wins is that you have more control on what is exported and a healing process, something which in practice turns out to be a useful for data-integrity towards third parties. Further, note, that if the building blocks are re-used an extra configuration will be needed.
- I am aware there are still loose ends in all services (e.g. naming), but the basics are covered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,3 +12,37 @@ Backend for the Contact Hub application based on the  mu.semte.ch microservices 
 You can shut down using `docker-compose stop` and remove everything using `docker-compose rm`.
 
 
+### Setting up the delta-producers related services
+To make sure the app can share data, producers need to be set up. There is an intial sync, that is potentially very expensive, and must be started manually
+
+#### producers administrative-units
+
+1. make sure the app is up and running, the migrations have run
+2. in docker-compose.override.yml, make sure the following configuration is provided:
+```
+  delta-producer-pub-graph-maintainer-administrative-units:
+    environment:
+      START_INITIAL_SYNC: 'true'
+```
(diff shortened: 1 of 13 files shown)


nbittich:
Looks 👍🏼  to me, I will let @bdevloed looking at it before merging

bdevloed (approved):

bdevloed:
Converted to draft due to merge conflicts and wip-code (orgs)

bdevloed:
Added a refactored Kalliope API to use the delta producer

nbittich (approved):

PR #69, opened 2021-11-23
migrate private data

nbittich:


bdevloed (approved):
Looks good, thx!
I've ?s before the optional because it sometimes creates counter intuitive results when unbound variables appear in sequential optional statements. (not a big issue here though)

PR #70, opened 2021-11-23
Add a new `/request-reasons` endpoint

Windvis:
This can be used to retrieve all the possible options.


config/migrations/20211119151700-move-reason-codes.sparql
```diff
@@ -0,0 +1,14 @@
+DELETE {
```
Windvis, line 1: These were inserted in the default graph by the migrations service so mu-cl-resources didn't return them.

nbittich (approved):

PR #71, opened 2021-11-23
Remove config related to the sensitive data feature

Windvis:
I forgot to remove this in the previous PR.


nbittich:
I also removed the config to forbid the sparql endpoint in prod, can u check? if ok can be merged

PR #75, opened 2021-12-01
Change event model updates

Windvis:
This provides the needed backend changes (and some mock data) for the change event functionality.

You can mount the mock data in your docker-compose.override.yml file.
```yml
  migrations:
    volumes:
      - ./config/dev-migrations:/data/migrations/dev
```

### DIFF
--- a/config/dev-migrations/20211209101400-mock-change-events.graph
+++ b/config/dev-migrations/20211209101400-mock-change-events.graph
@@ -1,0 +1,1 @@
+http://mu.semte.ch/graphs/organisatieportaal
--- a/config/resources/change-events.json
+++ b/config/resources/change-events.json
@@ -8,7 +8,8 @@
     "ext": "http://mu.semte.ch/vocabularies/ext/",
     "org": "http://www.w3.org/ns/org#",
     "regorg": "http://www.w3.org/ns/regorg#",
-    "skos": "http://www.w3.org/2004/02/skos/core#"
+    "skos": "http://www.w3.org/2004/02/skos/core#",
+    "foaf": "http://xmlns.com/foaf/0.1/"
   },
   "resources": {
     "change-events": {
@@ -43,6 +44,11 @@
           "predicate": "code:veranderingsgebeurtenisResultaat",
           "target": "change-event-results",
           "cardinality": "many"
+        },
+        "decision": {
+          "predicate": "http://data.europa.eu/m8g/hasFormalFramework",
+          "target": "decisions",
+          "cardinality": "one"
         }
       },
       "new-resource-base": "http://data.lblod.info/id/veranderingsgebeurtenissen/"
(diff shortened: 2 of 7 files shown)


config/dev-migrations/20211209101400-mock-change-events.graph
```diff
@@ -0,0 +1 @@
+http://mu.semte.ch/graphs/contacthub/141d9d6b-54af-4d17-b313-8d1c30bc3f5b/ChAdmin
```
Windvis, line 1: Not sure if this is still the correct graph after the role changes

config/resources/change-events.json
```diff
@@ -0,0 +1,104 @@
+{
```
Windvis, line 1: I moved it into a separate file, but I'm not sure if this is the best way to organize it since "decisions" aren't only related to the change events.

config/resources/change-events.json
```diff
@@ -0,0 +1,104 @@
...
+    "decisions": {
+      "class": "http://data.vlaanderen.be/ns/besluit#Besluit",
+      "attributes": {
+        "publication-date": {
+          "type": "datetime",
+          "predicate": "http://data.europa.eu/eli/ontology#date_publication"
+        },
+        "document-link": {
+          "type": "url",
+          "predicate": "foaf:page"
```
Windvis, line 98: I went with `foaf:page` for now.

config/resources/domain.json
```diff
@@ -299,14 +299,22 @@
...
         },
         "resulted-from": {
-          "predicate": "org:resultedFrom",
+          "predicate": "org:resultingOrganization",
           "target": "change-events",
-          "cardinality": "many"
+          "cardinality": "many",
+          "inverse": true
+        },
+        "changes": {
```
Windvis, line 313: probably not the most clear relationship name. Any suggestions?

config/dev-migrations/20211209101400-mock-change-events.graph
bdevloed: Indeed, should be http://mu.semte.ch/graphs/organisatieportaal

config/resources/change-events.json
bdevloed: The domain.json is getting a bit too big, so I think it's good to split it up.
I guess we'll find the right granularity through trial and error. So far I think it's sensible to have the decisions in here. Splitting things up per class could be a bit fragmented.

Windvis:
@nbittich Does this also require changes to the mu-auth config? I'm not sure how that service works exactly.

nbittich:
> @nbittich Does this also require changes to the mu-auth config? I'm not sure how that service works exactly.

You might need to add `http://lblod.data.gift/vocabularies/organisatie/VeranderingsgebeurtenisResultaat` in the `@protected_resource_type` array in `config.ex`

config/resources/domain.json
bdevloed: Why not changeEvents?

bdevloed (approved):

config/resources/domain.json
Windvis: It is linked to the `ChangeEventResults` so `changeEventResults` would be more correct, but that seemed to long / specific so I simplified it to `changes`. But I can go for that one if you prefer.

bdevloed: I would go for clarity, i.e. `changeEventResults`

nbittich:
is it ready to be merged? Looks good to me but I see Boris didn't approve it yet. If so I will merge it

Windvis:
@nbittich I think so. Everything seemed to work as expected in the frontend so feel free to merge if you think it's fine as well. :+1:

PR #76, opened 2021-12-03
acm idm roles

nbittich:
- adding mock login users with 3 types of roles (editor,reader,admin)
- refactoring mu-auth config to take roles into account
- use the standard acm idm login image


bdevloed:
Something is still not ok I think.
I get no mock users when following these steps:
- start from scratch on a development branch
- spin up docker compose (incl dev) and wait for migrations to run
- bring down docker compose
- switch to this branch
- spin up docker compose again (incl dev) and wait for migrations to run
- check http://localhost/mock-login
I'll look further into it next week

nbittich:
@bdevloed  strange but not impossible :s @Windvis  could you try it to see if you have the same issue ? thanks

nbittich:
nvm, orders in the migrations were wrong:
```
you:
migrations_1                                                | /data/migrations/2022212411511111121112333225-mock-user-roles.sparql [DONE]
migrations_1                                                | /data/migrations/2022212411511111121112333225-clear-mock-data.sparql [DONE]
me:
migrations_1                                                | /data/migrations/2022212411511111121112333225-clear-mock-data.sparql [DONE]
migrations_1                                                | /data/migrations/2022212411511111121112333225-mock-user-roles.sparql [DONE]
```

bdevloed (approved):

Windvis:
I know it's already merged, but I find the naming of these migration files confusing. We should be using a unix timestamp of the date the migrations are created, right? This seems to not be the case here which might impact migrations we create in the future?

nbittich:
> I know it's already merged, but I find the naming of these migration files confusing. We should be using a unix timestamp of the date the migrations are created, right? This seems to not be the case here which might impact migrations we create in the future?

sorry completely forgot to reply. It was at the time when we had the private-data folder and the timestamp were not known / buggy (as data were private and not everyone in the teams had accesses to them). We needed the mocks migration to be done after the private data related migrations.  it should be indeed a timestamp, maybe we should try to figure out how to fix this in the future

PR #77, opened 2021-12-08
Fix/type eredienst export

bdevloed:


bdevloed:
This PR adds a missing attribute for local and central worship services and also fixes a typo in the config which was overlooked during a code review.

nbittich (approved):

PR #86, opened 2022-01-06
Use `date` instead of `datetime` were needed

Windvis:
If only the date is relevant we should not store it as a `datetime` since that would add timezone information which might cause the date to be displayed incorrectly in the frontend.


Windvis:
Not sure if this is needed, but I noticed that dates which were persisted as `2021-11-03` would get returned as `2021-11-03T00:00:00` afterwards.

nbittich (approved):

PR #88, opened 2022-01-10
add contact types

bdevloed:
contact points as secondary when:
- There is NO address linked to the contactpoint AND another contact point for the mandataris exists.

Label all other contactpoints as primary. i.e. 
- only contact point (with or without address)
- OR the contact point with an address in case of multiple contact points per mandataris.

fyi, there are no cases of more than one contactpoint per mandataris with an address.

### DIFF
--- a/config/migrations/2022/20220202120000-primary-contacts.sparql
+++ b/config/migrations/2022/20220202120000-primary-contacts.sparql
@@ -1,7 +1,6 @@
 # cf 20220110100000-secondary-contacts.sparql for context.
 # In case it's not a secondary contactpoint (a contactpoint without an address if multiple contact points exist for one entity/mandataris)
 PREFIX locn: <http://www.w3.org/ns/locn#>
-PREFIX shema: <http://schema.org/>
 PREFIX schema: <http://schema.org/>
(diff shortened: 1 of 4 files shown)


Windvis:
Sounds good. I'm not the best to review the SPARQL query itself though. 

Should we wait to merge this until the frontend code uses the same logic? Otherwise we might have to run the migration again in the future?

bdevloed:
It can be done before I think. This migration only adds information which isn't used yet. No existing triples are deleted or modified. 
But let me know if I miss something.

Windvis:
Maybe I'm the one misunderstanding. What I believe this does is properly "label" the contact points. So instead of depending on the order in the database (and assume the first one is primary) we can use the `Primary` and `Secondary` labels instead?

If we run this migration now it will update the database to this new system, but the frontend will still create new contact points without these labels. So unless we also implement the frontend changes before merging I think there will be a period were "wrong" data can be inserted into the DB? So we would have to run this a second time when the actual frontend work is done?

bdevloed:
@Windvis  ah yes indeed. Will put it in draft in the meantime to avoid an accidental merge.

PR #100, opened 2022-02-10
improve perf bestuur

nbittich:


claire-lovisa (approved):
Tested and works as advertised ! ;)

PR #105, opened 2022-02-28
List all positions for a person (ministers + mandatories) & bug fix editing issue with delta 

nbittich:


claire-lovisa (approved):
Top :dancers:

PR #107, opened 2022-03-02
reschedule op dump creation

bdevloed:
Rescheduled to optimize KAL integration (at minute 45 every hr)


nbittich (commented):
this seems to be every 42 seconds of every minutes?

claire-lovisa:
> this seems to be every 42 seconds of every minutes?

no because there is no cron variable for the seconds so it goes straight to minutes !

claire-lovisa:
> I think it should be `CRON_PATTERN_DUMP_JOB: '0 45 * * *'`

Both work (i tested to be sure haha) ! Depends on the languages though, the cron job library of JS is less flexible.

claire-lovisa (approved):

bdevloed:
Thanks for the review :)
I've updated the comment.
Minute 42 was chosen to have a bit of buffer for the graph to be up to date (after last changes at minute 40) and the dump being ready before the pull scheduled at minute 45 by KAL.

PR #110, opened 2022-03-04
Add migration to delete CKB Wachtebeke

claire-lovisa:


nbittich (approved):
I didn't test it but from what I see it looks good to me!

PR #114, opened 2022-03-08
Remove werkingsgebieden from the config and db

claire-lovisa:


nbittich (approved):
tested locally, works as expected 👍🏼

PR #116, opened 2022-03-09
fix index

nbittich:
How to test:
- delete data/elasticsearch (sorry 😬 )
- to make testing faster, add the following in config.json (below number_of_threads):
    - `"max_batches": 1,`
- wait for the indexing to happen
- use any mock login user and check that it doesn't reindex again


claire-lovisa (approved):
No reindexing :partying_face:

PR #121, opened 2022-03-22
fix relationship ws

nbittich:


nbittich:
bugfix for [OP-1236](https://binnenland.atlassian.net/browse/OP-1236)

claire-lovisa (approved):

PR #128, opened 2022-04-07
fix regression mu search

nbittich:
bugfix for OP-1284. 

No reindex needed, just test if you can still edit a person & a bestuur


claire-lovisa (approved):
Works fine with persons & administrative units update, as well as to trigger a single reindex on a specific data !

PR #157, opened 2022-06-13
normalize spaces

nbittich:
OP-1495:

For some reasons there are few remainings after script runs, but it should be just a few.
```sparql
select ?o ?normalize  where {
  graph <http://mu.semte.ch/graphs/organisatieportaal> {
    ?s ?p ?o.
    BIND (replace(str(?o), "^(\\s+)|(\\s+)$", "") as ?removeTrailingSpace).
    BIND (replace(?removeTrailingSpace, "\\s{2,}", " ") as ?normalize).

    filter(isLITERAL(?o) && (regex(?o, "\\s{2,}") || regex(?o, "^(\\s+)|(\\s+)$")) )

  } 
}
```

### DIFF
--- a/config/migrations/2022/202206131O4623-normalize-space.sparql
+++ b/config/migrations/2022/202206131O4623-normalize-space.sparql
@@ -12,8 +12,9 @@ INSERT {
 WHERE {
   graph <http://mu.semte.ch/graphs/organisatieportaal> {
     ?s ?p ?o.
-    BIND (replace(?o, "^(\\s+)|(\\s+)$", "") as ?removeTrailingSpace).
-    BIND (replace(?removeTrailingSpace, "\\s{2,}", " ") as ?normalize).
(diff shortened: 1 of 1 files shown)


nbittich:
@claire-lovisa  I've updated the query, while it seems working now, could you double check?

I've updated the query to take tabs into account as well:

```sparql
select ?o ?normalize  where {
  graph <http://mu.semte.ch/graphs/organisatieportaal> {
    ?s ?p ?o.
    BIND (replace(str(?o), "^(\\s+)|(\\s+)$|(\\t)$|(\\t)", "") as ?removeTrailingSpace).
    BIND (replace(?removeTrailingSpace, "\\s+|\\t", " ") as ?normalize).

    filter(isLITERAL(?o) && (regex(?o, "\\s+|\\t") || regex(?o, "(\\s+)|(\\s+)$|(\\t)$|(\\t)")) )

  } 
}

claire-lovisa (approved):
Ah tabs, that was the culprit !! Seems good now :dancers:

PR #171, opened 2022-09-19
duplicate municipality for aalter

nbittich:
OP-1775, unfortunately I couldn't do it in one query :(
For the elasticsearch reindexing, it just needs to "bewerk" then "opslaan" each aalter after the migrations ran, and it should trigger reindexing. I asked Sofie if she could do it, as there are only 3 with the problem.
I  struggled with delta message and curly curl today, that wasn't my day


nbittich:
@claire-lovisa the address is linked to the site `Http://data.lblod.info/id/vestigingen/7c6b24aac9de9c25aec3a1b3962235f6` and the three units share the same primary site in the migrations

claire-lovisa:
@nbittich And isn't this weird ? I'd expect we want to be able to modify them separately :thinking:

nbittich:
> @nbittich And isn't this weird ? I'd expect we want to be able to modify them separately thinking
ah that I don't know :D we should check wit @bdevloed

PR #173, opened 2022-09-27
OP-1812, OP-1813, OP-1814 + lang typed literals

bdevloed:


nbittich (approved):
It will require a full reindex to test it in the fronted; when deployed the procedure is to stop containers elasticsearch & search, delete data/elasticsearch directory and restart the sopped container

PR #185, opened 2022-10-20
Fix/op 1806

bdevloed:
https://binnenland.atlassian.net/browse/OP-1806


claire-lovisa (approved):
Tested, seems to be removing the duplicates :)

PR #187, opened 2022-10-21
OP-1708 Fix typo in name

iurianu:


claire-lovisa (changes requested):
Hi ! When testing the migrations I saw that the frontend wasn't updated. When I looked a bit deeper, I noticed you update the wrong URI, it should be `<http://data.lblod.info/id/representatieveOrganen/0ebb44c2ef14d86978ea85e74d128ad1>`. How did you get `<http://lblod.data.gift/concepts/0ebb44c2ef14d86978ea85e74d128ad1>` ?

PR #188, opened 2022-10-27
Experiment/mu search

nbittich:
- use the main repo for mu search
- use cluster mode (not really useful in case of OP, as we have just a few indexes)
- reduce memory consumed by elasticsearch instance(s)

I tested it locally and it works (didn't see big improvement in speed though, but my local machine is super slow)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ docker-compose.override.yml
 .idea
 *.iml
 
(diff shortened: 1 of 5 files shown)


claire-lovisa:
I gave it a try and it seems to be working nicely ! Didn't see any weird behaviors :)

PR #189, opened 2022-10-27
Add construct-administrative-unit-relationships service and config

claire-lovisa:
OP-1696

Has to be tested with https://github.com/lblod/frontend-organization-portal/pull/414 and https://github.com/lblod/construct-administrative-unit-relationships-service/pull/1


bdevloed (approved):
Works and code is easy to read :)

PR #190, opened 2022-11-16
Fix / Create dummy worship service 'Kerkfabriek Vredegem' in OP

iurianu:
OP-1891


iurianu:
I found the code piece, in Felix's update task for Loket (DL-4440)

claire-lovisa (changes requested):
Ah, it's actually a little more complicated than expected. When I tested the code, so when I tried to visit the bestturseenheid page, it break. While looking why, I saw that in a recent migration the entire entity has been deleted [see here](https://github.com/lblod/app-organization-portal/blob/development/config/migrations/2022/20221015110500-delete-kerkfabriek-st-baafs.sparql). So we need to add it back, from the data dump I created I see that it looked like this but you'll have to double check that it works properly : 
```
<http://data.lblod.info/id/besturenVanDeEredienst/578bf96f86056e2260dc21cee91aa519> a ere:BestuurVanDeEredienst,
        besluit:Bestuurseenheid ;
    ere:typeEredienst <http://lblod.data.gift/concepts/b13d1d623626bc1ee75c7d20bc60e3c0> ;
    besluit:classificatie <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/66ec74fd-8cfc-4e16-99c6-350b35012e86> ;
    mu:uuid "578bf96f86056e2260dc21cee91aa519" ;
    skos:prefLabel "Kerkfabriek St.-Baafs van Gent" ;
    regorg:orgStatus <http://lblod.data.gift/concepts/63cc561de9188d64ba5840a42ae8f0d6> .
```

Once this is done, we will need to trigger the indexing for this specific bestuur manually, to avoid having to do a complete indexing that takes a lot of time. The call will look somewhat like below, but you have to test to see if it does what we want (= being able to search for the bestuur in the main page and find it)
```
curl --location --request POST '<ip-of-mu-search>/update' \
--data-raw '    [
      {
          "deletes": [
              {
                  "subject": { "type": "uri", "value": "http://data.lblod.info/id/besturenVanDeEredienst/578bf96f86056e2260dc21cee91aa519" },
                  "predicate": { "type": "uri", "value": "http://www.w3.org/2004/02/skos/core#prefLabel" },
                  "object": { "type": "string", "value": "Kerkfabriek St.-Baafs van Gent" }
              }
          ],
          "inserts": [
              {
                  "subject": { "type": "uri", "value": "http://data.lblod.info/id/besturenVanDeEredienst/578bf96f86056e2260dc21cee91aa519" },
                  "predicate": { "type": "uri", "value": "http://www.w3.org/2004/02/skos/core#prefLabel" },
                  "object": { "type": "string", "value": "Kerkfabriek Vredegem" }
              }
          ]
      }
    ]'
```

claire-lovisa:
It gets added to the wrong graph, when you add a `.ttl` file to upload data, you also need to create a file with the same name but with extension `.graph` with the graph name in it, so `http://mu.semte.ch/graphs/organisatieportaal` for OP

PR #199, opened 2022-12-07
Fix/incidents-handling

iurianu:
OP-1928 | OP-1929 | OP-1954

### DIFF
--- a/config/migrations/2022/20221206100001-delete-duplicate-persons-st-franciscus-glabbeek.sparql
+++ b/config/migrations/2022/20221206100001-delete-duplicate-persons-st-franciscus-glabbeek.sparql
@@ -1,0 +1,61 @@
+# OP-1928 : Remove worship members in "Kerkfabriek St.-Franciscus van Glabbeek"
+
+# Delete duplicate persons
+DELETE {
+  GRAPH <http://mu.semte.ch/graphs/organisatieportaal> {
+    <http://data.lblod.info/id/personen/635BD6EC8DE5818A7C9F4310> ?pm ?om .
+    <http://data.lblod.info/id/personen/635BD7518DE5818A7C9F4313> ?pj ?oj .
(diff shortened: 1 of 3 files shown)


claire-lovisa (changes requested):
The migration works well and the data is updated as expected in the triplestore, but we are missing a curl call to update the index on the `personen` page : we can still see the duplicates there !

iurianu:
**Curl Call to update the index:**

curl --location --request POST '172.25.0.17/update' \
--data-raw '    [
  {
    "deletes": [
      {
        "subject": { "type": "uri", "value": "http://data.lblod.info/id/personen/635BD6EC8DE5818A7C9F4310" },
        "predicate": { "type": "uri", "value": "http://mu.semte.ch/vocabularies/core/uuid" },
        "object": { "type": "string", "value": "635BD6EC8DE5818A7C9F4310" }
      },
      {
        "subject": { "type": "uri", "value": "http://data.lblod.info/id/personen/635BD7518DE5818A7C9F4313" },
        "predicate": { "type": "uri", "value": "http://mu.semte.ch/vocabularies/core/uuid" },
        "object": { "type": "uri", "value": "635BD7518DE5818A7C9F4313" }
      }
    ],
    "inserts": []
  }
]'

claire-lovisa:
@iurianu could you please remove the names of the persons from the commit message and from the files ?   For their privacy 😅

iurianu:
> @iurianu could you please remove the names of the persons from the commit message and from the files ? For their privacy sweat_smile
sure, no problem!

claire-lovisa:
Two remarks on the second migration : 
1. I would have kept the other person's URI because the other one had the gender, this one has nothing
2. In the curl call you'll need to remove the extra positions for the person that you keep. You can see that when typing the name of the person in the search for the Personen page, it still shows up multiple times

claire-lovisa (approved):

PR #206, opened 2022-12-27
Fix/ Incident Handling OP

iurianu:
OP-1977 | OP-1978 | OP-1979 | OP-1980 | OP-1981


claire-lovisa (changes requested):
Some data is left after this deletion, for example the identifier, the site and its address ... The removing needs to go deeper, see previous migrations where it has been done

nbittich:
It might not be enough, there might be orphan persons left behind (see https://github.com/lblod/app-organization-portal/pull/208/files)

If there are persons with positions in those bestuur (and not anywhere else), you might need to either migrate them to the correct bestuur or delete them completely

PR #207, opened 2023-01-03
fix missing persons & use standard mu search

nbittich:
OP-1987

requires a full reindex (sorry)
requires https://github.com/lblod/frontend-organization-portal/pull/430


claire-lovisa (approved):
Tested and it works fine + the bugs is resolved ! Not sure how, I'm assuming that bumping mu-search was key, but anyways it works :ok_hand: :smile:

PR #211, opened 2023-01-10
apply worship-app changes

bdevloed:
use same consumer logic as the worship app
- updated mapping rules
- use lblod consumers
- updated reasoner image


bdevloed:
Query to check which data is in the privacy sensitive graph:
```
SELECT DISTINCT ?domain ?attribute ?range (COUNT(?s) as ?numberOfConnections) WHERE {
  GRAPH <http://mu.semte.ch/graphs/privacy-centric-graph> {
    ?s
      ?attribute
        ?o.
    OPTIONAL {
      ?s
        a
          ?domain
    }
    {
      ?o
        a
          ?range.
    }
    UNION
    {
      ?s
        ?attribute
          ?o.
      BIND(datatype(?o) as ?range)
    }
    FILTER(bound(?range))
  }
} 
GROUP BY ?domain ?attribute ?range
ORDER BY ?domain
```

Verfiy which privacy sensitive info might be in the main graph:
```
select distinct ?s ?p ?o {
  GRAPH <http://mu.semte.ch/graphs/privacy-centric-graph> {
    ?s ?p ?o
  }

  GRAPH <http://mu.semte.ch/graphs/organisatieportaal> {
    ?s ?p ?o 
  }
}
```

config/reasoner/dl2op/main-info/rrn.n3
```diff
@@ -0,0 +1,62 @@
...
+    {
+      ?identifier skos:notation ?notation.
+    } e:becomes {
+      ?identifier
+          generiek:gestructureerdeIdentificator ?structuredIdentifier ;
+          skos:notation "Rijksregisternummer" .
+    }.
+
+} => {
+  ?identifier
```
claire-lovisa, line 44: Here I think that the type is missing, at least I don't see it in the db when I consume it !
( `http://www.w3.org/ns/adms#Identifier` )

config/reasoner/dl2op/private-info/rrn.n3
```diff
@@ -0,0 +1,62 @@
...
+    {
+      ?identifier skos:notation ?notation.
+    } e:becomes {
+      ?identifier
+          generiek:gestructureerdeIdentificator ?structuredIdentifier ;
+          skos:notation "Rijksregisternummer" .
+    }.
+
+} => {
+  ?identifier
```
claire-lovisa, line 44: Or here, for the type ? Not sure why it's in the main-info config as it should only end up in the privacy graph, but i'll let you put it in the correct place haha

claire-lovisa (changes requested):

config/reasoner/dl2op/private-info/rrn.n3
bdevloed: I choose to have the exact same rules in all the configs. So the queries - i.e. filters. Can be done using the OP domain language.
- this eases the pain of having to thing how the stuff we don't want in OP is modelled in Loket. It will also help us once Loket fixes a few things. We don't need to change anything, we could clean up the rules that are no longer triggered at our own pace. Everything will keep working as expected
- It's also simple - for me at least - to only have one kind of mapping logic. The performance impact is negligible. Moving and parsing data takes much more time than the actual reasoning.

config/reasoner/dl2op/main-info/rrn.n3
bdevloed: Correct. Well spotted :) I've now added it to the private-info query.

PR #213, opened 2023-01-17
Update migration to avoid data loss

claire-lovisa:
The migration was removing all the contact points, all the addresses and all the identifiers of the database.

It's important to order the optionals correctly, each optional is a distinct block and what happens in an optional block stays in the optional block. So a variable defined in an optional block can't be reused outside of this block.


nbittich (approved):
good catch!

PR #214, opened 2023-01-18
bugfix non utf8 characters

nbittich:
OP-2044 && OP-1915

would be nice to have it for 1.14

### DIFF
--- a/config/migrations/2023/20230118093645-fix-non-utf8.sparql
+++ b/config/migrations/2023/20230118093645-fix-non-utf8.sparql
@@ -1,0 +1,25 @@
+delete {
+graph <http://mu.semte.ch/graphs/organisatieportaal> {
+    ?s <http://xmlns.com/foaf/0.1/givenName> ?givenName.
+  }
+}
+
+insert {
+graph <http://mu.semte.ch/graphs/organisatieportaal> {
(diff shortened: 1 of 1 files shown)


claire-lovisa (approved):
Do you also have a curl call, to update QA's index without having to reindex it ?
No problem to ship it with 1.14, you can merge directly to master then, I have a frontend fix to do as well, i'll deploy both at the same time on QA, then prod when it's all ok :)

bdevloed (approved):
Looks good to me. I guess what's still pending - in other prs - is to:
- find the root cause
- verify all strings, not just `André`?

nbittich:
> Looks good to me. I guess what's still pending - in other prs - is to:
> 
>     * find the root cause
> 
>     * verify all strings, not just `André`?

I verified all ?o with type xsd:string  (added csv file under OP-1915 the git sparql.csv)

seems that's all we have. iconv gives 0 as an output.

nbittich:
> Do you also have a curl call, to update QA's index without having to reindex it ? No problem to ship it with 1.14, you can merge directly to master then, I have a frontend fix to do as well, i'll deploy both at the same time on QA, then prod when it's all ok :)

I've added the curl calls

PR #215, opened 2023-01-18
Fix / migration to delete bedienaren

iurianu:
OP-2043 - Delete "Bedienaren" from worship service in Production.
OP-2047 - Delete a bedienaar from OP in Production


claire-lovisa (changes requested):
I see two issues here:
- The positions have contact points attached, they also need to be deleted
- I don't get why you say that a curl call isn't needed, as you can see in the screenshot below the deleted positions still appear in the Personen view ?

Capture d’écran du 2023-01-18 18-15-32

claire-lovisa (changes requested):
The contact points have addresses that also need to be deleted. You should check all the data you're removing before creating the PR, to ensure you remove all the necessary parts.

The curl calls are not working, the subjects should be predicates, the predicates should be objects and the objects should be subjects. I suggest that you first setup the app so that you can see the persons in the `Personen` page, then test the curl calls to ensure they work before asking for review.

config/migrations/2023/20230120150000-delete-bedienaren-from-worship-service.sparql
```diff
@@ -0,0 +1,75 @@
...
+PREFIX ch: <http://data.lblod.info/vocabularies/contacthub/>
+PREFIX ere: <http://data.lblod.info/vocabularies/erediensten/>
+
+DELETE {
+  GRAPH <http://mu.semte.ch/graphs/organisatieportaal> {     
+    ?rolBedienaar ?pRolBedienaar ?oRolBedienaar .
+    ?contactPoint ?pContactPoint ?ocontactPoint .
+    ?address ?pAddress ?oAddress .
+    ?positieBedienaar ?pPositieBedienaar ?oPositieBedienaar .
+    ?functie ?pFunctie ?oFunctie .
```
claire-lovisa, line 14: Those two should not be deleted, they are shared with other positions

claire-lovisa (changes requested):

claire-lovisa:
I also think that the curl call should be as followed, because it doesn't work with the current one
```
curl --location --request POST '172.19.0.30/update' \
--data-raw ' [
  {
    "deletes": [
      {
        "subject": { "type": "uri", "value": "http://data.lblod.info/id/rollenBedienaar/6332F770036B579F0A26FF2F" },
        "predicate": { "type": "uri", "value": "http://www.w3.org/ns/org#heldBy" },
        "object": { "type": "uri", "value": "http://data.lblod.info/id/personen/6332F6FD036B579F0A26FF2B" }
      },
      {
        "subject": { "type": "uri", "value": "http://data.lblod.info/id/rollenBedienaar/6332F6C9036B579F0A26FF2A" },
        "predicate": { "type": "uri", "value": "http://www.w3.org/ns/org#heldBy" },
        "object": { "type": "uri", "value": "http://data.lblod.info/id/personen/6332F648036B579F0A26FF26" }
      },
      {
        "subject": { "type": "uri", "value": "http://data.lblod.info/id/rollenBedienaar/6332F7FB036B579F0A26FF34" },
        "predicate": { "type": "uri", "value": "http://www.w3.org/ns/org#heldBy" },
        "object": { "type": "uri", "value": "http://data.lblod.info/id/personen/6332F7AF036B579F0A26FF30" }
      },
      {
        "subject": { "type": "uri", "value": "http://data.lblod.info/id/rollenBedienaar/638F083FA742F4FD4AB55907" },
        "predicate": { "type": "uri", "value": "http://www.w3.org/ns/org#heldBy" },
        "object": { "type": "uri", "value": "http://data.lblod.info/id/personen/638F07A6A742F4FD4AB55903" }
      }
    ],
    "inserts": []
  }
]'
```

bdevloed:
@iurianu / @claire-lovisa migration and curl work as expected. The positions are indeed removed.

PR #216, opened 2023-01-18
delete empty literals

bdevloed:
OP-1329

No new empty literals should be created
- frontend has been adapted (OP-1331)
- loket consumers retract empty literals

No re-index triggers needed


claire-lovisa (approved):
:partying_face:

PR #217, opened 2023-01-18
Fix / Change bedienaar roles

iurianu:
OP-2045 | OP-2046 | OP-2056 | OP-2057 | OP-2060


claire-lovisa (approved):
The migration does the changes requested :+1: 
But I think a curl call is also needed here, see screenshot, to get the position name right
image

claire-lovisa (approved):

PR #219, opened 2023-01-19
use same graph for dcat and files to avoid bug

bdevloed:
cf OP-2069

Work around bug in dump-file publisher when deleting old dumps https://github.com/lblod/delta-producer-dump-file-publisher/pull/6


claire-lovisa (approved):
Does it require to migrate previous files info ? To put them in this new graph ? Or will it get fixed by itself when a new dump file is created ?

bdevloed:
Indeed, in it does. But I think it's best to clean up everything instead. As in, the history will not be kept after this update anyway.
So I left out the migration I wrote before. I can still add it though.

nbittich (approved):

claire-lovisa:
nice :ok_hand:

nbittich (approved):

PR #226, opened 2023-01-25
migration to delete mandatarissen, and person

nbittich:
OP-2073

I've recreated the branch from @iurianu  and cherry picked his commit since the previous branch was based on development branch instead of master


claire-lovisa (changes requested):
The mandatarissen have contact points that need to be deleted, and the contact points have addresses that also need to be cleaned up.

claire-lovisa:
Also the curl call isn't working, maybe try targetting the relationship between the person and the mandataris ?

claire-lovisa (approved):

PR #229, opened 2023-01-28
mu search delta script

nbittich:
history is kept in a file and not in db. this file is git ignored.


claire-lovisa (approved):
Tested and approved :smile:

PR #231, opened 2023-01-31
Migration to remove region

iurianu:
OP-1994 | OP-2033


claire-lovisa (changes requested):
You're removing too much data, the locations should remain. The current migration for example also removes the region of multiple other municipalities, such as Gemeente Laarne. I think in this case you only need to delete the link `?werkingsgebiedgeo:sfWithin ?locatie`, not the locatie itself

claire-lovisa (approved):
Can you merge this to master ? Thanks !

PR #234, opened 2023-02-03
remove dummy worship service

iurianu:
OP-2125


claire-lovisa (commented):
Did the curl call work for you ? Doesn't seem to do its job when I try it

iurianu:
> Did the curl call work for you ? Doesn't seem to do its job when I try it

claire-lovisa:
It still doesn't work for me. Could some else take a look ? (cc @bdevloed haha)
I've been also trying to use the new migrations version here https://github.com/lblod/app-organization-portal/pull/235 and it seemed to work, and in there no need for curl calls, so we could also go with that option

claire-lovisa:
Closing, I'll merge #235 which contains the fix of this PR

PR #235, opened 2023-02-06
Migrations triggering reindex

claire-lovisa:
OP-2117

I used a [currently ongoing PR](https://github.com/lblod/app-organization-portal/pull/234) to test wiring mu-migrations to mu-auth to see how would the reindexing work in our stack. And as a first test, it looks conclusive !

I wired it as a separate service because our previous migrations are full of comments and maybe other things that mu-auth does not like. So if we would go to having just one migrations service, we'd have to also clean older migrations to ensure that the project would be able to run from an empty DB. And I think that it doesn't hurt to keep it as a second service for now, gives us more time to try it out and see how the auto-reindexing goes.

It also means that now, before starting `migrations-triggering-indexing`, some services should be up for the changes to be picked up (triplestore, db, deltanotifier, search, elasticsearch)

### DIFF
--- a/config/migrations-triggering-indexing/2023/20230203150000-remove-dummy-worship.sparql
+++ b/config/migrations-triggering-indexing/2023/20230203150000-remove-dummy-worship.sparql
@@ -48,20 +48,3 @@ DELETE {
     }        
   }
 }
-
-# The associated call to trigger a reindex is :
-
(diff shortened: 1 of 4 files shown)


nbittich (approved):
works as expected 👍🏼

PR #236, opened 2023-02-13
add new codelist

nbittich:
OP-2139


claire-lovisa (changes requested):
As OP is the master of the minister positions (`ere:PositieBedienaar`), we also need to update the logic to create the necessary minister positions on admin unit creation [here](https://github.com/lblod/construct-administrative-unit-relationships-service/blob/master/lib/config.js#L297) I think ?

And we should also probably, for each EB, create a new minister position of type `Waarnemende imam` and export them to Loket ? Otherwise in Loket they will not have the option to select this position type

claire-lovisa (approved):

PR #239, opened 2023-02-20
Consume mandatarissen, leidinggevenden and worship positions from Loket

bdevloed:
Filter code hosted (for now) here: https://github.com/bdevloed/consumer-dispatcher

The initial sync should take about 10 mins. Incoming data is stored in the ingest-graph

Dispatching/filtering is based on the config: https://github.com/bdevloed/consumer-dispatcher/blob/master/dispatch-config.js

Inital version is only based on types. Not yet additional filter criteria


config/loket-consumer/leidinggevenden-dispatching/initial-sync-dispatching.js
```diff
@@ -36,20 +36,24 @@ async function dispatch(lib, data) {
...
-          endpoint,
-          BATCH_SIZE,
-          MAX_DB_RETRY_ATTEMPTS,
-          SLEEP_BETWEEN_BATCHES,
-          SLEEP_TIME_AFTER_FAILED_DB_OPERATION
-        )
-      }
+    let transformedInsertTriples;
+    try {
+      transformedInsertTriples = await transformStatements(fetch, originalInsertTriples);
```
claire-lovisa, line 41: This file is fine but I see that the delta sync file for leidinggevenden still contains .then(), it should be replaced by awaits to ensure the order of execution, especially as we do deletes and inserts :smile:

claire-lovisa:
I updated the code to:
- Adapt the migrations to the hardsplit
- Use the delta-consumer updates where it can now give context on received deltas

bdevloed:
@claire-lovisa I cannot formally approve this PR since I created it originally. But the updates/refactoring works as expected. :-)

claire-lovisa:
Thanks for taking the time to review and test @bdevloed ! :dancers: 

Waiting for the go of Sofie on OP-2264 before merging

PR #243, opened 2023-02-28
Feat/remove beheerder

nbittich:


claire-lovisa (approved):
The migrations renaming means they might re-run on dev ? Not sure if it's on purpose so i'm just mentioning it in case :sweat_smile:

nbittich:
@claire-lovisa  it is on purpose, otherwise they were set for 01/04/2023 😬

PR #244, opened 2023-02-28
Feature/extend public producer

bdevloed:
Extended for Loket
Saving wip here. will rebase and test further before getting it out of draft


claire-lovisa (approved):
I might have missed parts of it without the consumer, but from what I see it looks fine ! I am not sure why the commit "[Enable local involvements for central worship services](https://github.com/lblod/app-organization-portal/pull/244/commits/c5e1436ea8fed2e4ff609d22a41cc0ab9e86104e)" is there but it's already on prod so I guess it's OK :thinking:

PR #251, opened 2023-03-07
Incident/move change event

nbittich:
OP-2216


claire-lovisa (changes requested):
On top the the status update, we also need to change the resulting org no ? 
image

claire-lovisa (approved):

PR #256, opened 2023-03-21
Adapt public producer to hardsplit

claire-lovisa:
I'm PRing directly to master because it has been hotfixed to prod already and we need it in the next OP release.

I didn't changed the migrations as they will already have ran on prod.


config/authorization/config.ex
```diff
@@ -16,6 +16,7 @@ defmodule Acl.UserGroups.Config do
                         "http://data.lblod.info/vocabularies/erediensten/EredienstMandataris",
                         "http://data.lblod.info/vocabularies/erediensten/BestuurVanDeEredienst",
                         "http://data.lblod.info/vocabularies/erediensten/EredienstBestuurseenheid",
+                        "http://data.vlaanderen.be/ns/besluit#Bestuursorgaan",
```
nbittich, line 19: it is already on @org_type, this shouldn't be there I think. and it might produces issues (writing on multiple graph).

claire-lovisa: Ah ! Indeed !

nbittich (approved):

PR #260, opened 2023-03-31
migration from in rang to waarneemende

nbittich:
OP-2245


claire-lovisa (commented):
Hi :) I have a few comments:
- Why limiting the query to besturen that are "in oprichting" ? I don't see a mention on this restriction in the ticket ?
- The migration should trigger a reindex, to have the persons's search updated ? I ran it in the migrations-triggering-indexing folder, seemed to work fine

nbittich:
@claire-lovisa see OP-2242 but maybe I didn't understand correctly. I plan to do a full reindex for 1.16.1 as I don't trust enough the migration triggering indexing and we have private migrations that needs to run. It takes less time after the hard split and is safer in my opinion

claire-lovisa (approved):

PR #264, opened 2023-04-18
reset elastic script

nbittich:
[OP-1827]
docker & docker compose are needed, thus we cannot use mu scripts


claire-lovisa (approved):
Works perfectly :) Just a question, should we put the script it `./scripts/` instead of in `./config/scripts`, to keep them in the same place ? The mu-scripts are going under the `project` dir so they would still be separated

PR #268, opened 2023-05-02
 updating end date bestuursorgaan triggers too many deletes

nbittich:
OP-2307


claire-lovisa (approved):
Works fine, thanks for the fix ! I'm wondering why we had this weird behavior before, good to know that it's risky to define a type twice :thinking:

PR #269, opened 2023-05-03
Addition of ldes services to publish organization contact information

elpoelma:
This PR adds two new services:
- the `fragmentation-producer` (ldes-backend) service: this service is responsible to fragment and version incoming resources and publish them on a specified feed/stream.
- the `ldes-delta-pusher` service: this service takes incoming deltas, optionally filters them and sends them to a specified LDES backend.

In this application, these services are configured in order to version and fragment information and contact information of different organizations.

Known issues:
- each time an edit/creation of an organization occurs, it's data is sent too many times to the LDES backend.


config/ldes-delta-pusher/dispatch.ts
```diff
@@ -0,0 +1,59 @@
...
+		const subjects = new Set(changeset.inserts.map((insert) => insert.subject.value));
+		for (const subject of subjects) {
+			const { results: { bindings } } = await query(`
+        PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+        PREFIX code: <http://telegraphis.net/ontology/measurement/code#>
+        PREFIX org: <http://www.w3.org/ns/org#>
+        PREFIX adms: <http://www.w3.org/ns/adms#>
+  
+        CONSTRUCT {
+          ?organization a org:Organization;
```
claire-lovisa, line 16: I tested the PR by updating a phone number in a municipality. I didnt see any file generated, so I did a (very small) investigation, I found that the ldes-backend wasn't getting any activity, and when running manually the queries in Virtuoso I saw that enforcing the type `org:Organization` seemed to return nothing because Virtuoso isn't aware of inheritance. But if the query goes through mu-cl-resources it should I think, so I'm not sure why the backend doesn't receive anything. Can you describe the happy-path you use for testing ? Thank you :)

elpoelma: Yes I now see it doesn't return anything anymore while it did on the previous version. I could try specifying the necessary subtypes in the query.

elpoelma: So using the migration and for newly created organizations, organizations should have the org:Organization type. When creating/editing an organization and its contact info, they should be added to the `organizations` ldes-feed. The first page should be accessible through `http://localhost/ldes/organizations/1`

claire-lovisa (approved):
I've indeed been able to see some data written down with those adjustments ! :sparkles:

PR #272, opened 2023-05-03
Fix migration removing too much data

claire-lovisa:
OP-2326

This story of ministers, positions, ministers positions ... It's getting everybody confused, me included apparently :sweat_smile: 
In the previous migration, I removed the theoretical positions (`PositieBedienaar`) instead of the real position (aka the person in the position at a certain time, `RolBedienaar`).
It only got deployed on dev, so I restored the missing data on dev directly (using Loket as a source, as it was in sync). This migration should now only delete what's necessary, no more.

Deploy note: when deploying this fix to dev, it will also remove already-synced ministers as the initial sync has already run. We'll have to re-trigger an initial sync (should be fine as positions can't be edited in OP anymore). For QA and Prod,  no problem as the wrong migration didn't run and the sync isn't set up yet.


bdevloed (approved):

bdevloed:
It indeed really got me confused :D Thanks for the fix!

PR #275, opened 2023-05-09
feat/delete-dup-change-events

nbittich:
OP-2363


claire-lovisa (changes requested):
The change events also seem to have a decision linked via `?changeEvent <http://data.europa.eu/m8g/hasFormalFramework> ?decision`, it also needs to be deleted no ?

nbittich:
indeed, thank you @claire-lovisa  ! I've changed the migration, not sure if I should go more in depth with the relations of the decision (seems I shouldn't)

claire-lovisa (changes requested):
The decision isn't deleted :sweat_smile: I think it's because (i don't know why, that's weird but) the decision itself is in the `http://mu.semte.ch/graphs/administrative-unit` graph instead of in `http://mu.semte.ch/graphs/worship-service` ?

PR #282, opened 2023-06-07
sharepoint id for district, municipality, province & ocmw

nbittich:
OP-1864 & OP-1865


claire-lovisa (approved):
I didn't test them all but all the ones I opened had a sharepoint id :+1: 

One exception though for District Berchem it seems ?

PR #284, opened 2023-06-14
Leidinggevenden: Sync leidinggevenden from leidinggevenden databank

nbittich:
OP-2430

I think this is the only requirement for the sync, unless I misunderstood? thank you :)

### DIFF
--- a/config/reasoner/dl2op/main/query.n3q
+++ b/config/reasoner/dl2op/main/query.n3q
@@ -26,7 +26,7 @@
     ?p ?o .
 }.
 
-# Functionarissen should be linked to a gemeente, district, provincie or ocmw only (FOR NOW)
+# Functionarissen should be linked to a agb, apb, gemeente, district, provincie or ocmw only (FOR NOW)
 
 {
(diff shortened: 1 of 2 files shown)


bdevloed (approved):
Indeed :)

claire-lovisa:
The steps I took to test:
```
// Flushes all the sync jobs (initial sync and deltas) as well as the landing zone
curl --request POST <leidinggevenden-consumer-ip-address>/flush

// Re-starts an initial sync. No risk to ingest data that has been modified in OP as OP is read-only for the positions
drc restart leidinggevenden-consumer 

// Need to reindex to see the positions
sh ./scripts/reset-elastic.sh
```

What I noticed:
- We miss core data in OP (organen and bestursfuncties). It implies that even if we ingest the positions themselves, we don't have the necessary relationship to display the required information in the frontend
image
- I bumped the consumer, it shouldn't change how it behaves but it adds the flushing endpoint which is handy here

nbittich:
superseded by https://github.com/lblod/app-organization-portal/pull/293

PR #287, opened 2023-06-22
Added moslimraad - to be confirmed

bdevloed:
Address still to be confirmed (postal code, province)


bdevloed:
- postal code in request differs from KBO
- requested to add province, which is not correct for BXL.

bdevloed:
Conclusion, keep 1040 as postal code, and keep province to be consistent with other addresses in OP --> should be fixed later.

bdevloed:
will create new pr

PR #292, opened 2023-06-28
agb apb migration

nbittich:
OP-2422 && OP-2271, I kept some triples empty in order to avoid any *potential* regression in the frontend, and we should receive the up to date share point list with corrected address in the future.

I can remove them if it might cause issues with e.g synchronization


claire-lovisa (commented):
I think that even if we keep empty entities (so a structure with no meaningful info in it), we should at least try not to create empty literals (triples where the value is `""`).

From what I see in Jira, in OP-1328 we already got rid of empty literals no I don't really expect issues ..? And if it causes issues, it's probably that the frontend needs to be updated to handle it.

The work on empty entities (OP-2061) is still not done, so I'd not focus too much on that.

Other than that, the migrations executes properly and I see data so :+1: , maybe Boris will have a different opinion on the empty values !

nbittich:
@claire-lovisa  I removed the empty strings and also added ovo numbers (OP-2422)

claire-lovisa (approved):
Nice ! And I seem to be able to update the primary site address and contact info without errors, so I think it's safe not to insert the empty strings !

PR #298, opened 2023-07-06
added domain.json

bdevloed:
Adds the domain.

Some code lists still need to be added:

- sectors: based on currently used values, to be validated by business
- activities: subset of nacebel - to be defined, which subset

### DIFF
--- a/config/resources/domain.json
+++ b/config/resources/domain.json
@@ -382,7 +382,7 @@
           "type": "date",
           "predicate": "lblodgeneriek:geplandeEindDatum"
         },
-        "pupose": {
+        "purpose": {
           "type": "string",
           "predicate": "org:purpose"
         }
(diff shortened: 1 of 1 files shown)


config/resources/domain.json
```diff
@@ -374,6 +377,14 @@
         "alternative-name": {
           "type": "string",
           "predicate": "skos:altLabel"
+        },
+        "expected-end-date": {
+          "type": "date",
+          "predicate": "lblodgeneriek:geplandeEindDatum"
+        },
+        "pupose": {
```
lagartoverde, line 385: purpose, I guess :)

lagartoverde (changes requested):
Looks good, fix the typo and you can merge :)

config/resources/domain.json
bdevloed: 🙈 of course

PR #300, opened 2023-07-12
Add IGS positions to the sync

claire-lovisa:
OP-2513

For testing, it's a little bit tricky as we miss some data. I took some notes that you can reuse !

```
Test with leidinggevenden of Stadsregio Turnhout

Can not be tested the right way because we are missing too much data (primary sites data + labels of ISG)
So by adding this extra data, we'll at least be able to test for the leidinggevenden of ONE bestuur to see if the code
works or not. We add a label and an existing primary site, to make our lifes easier :)

INSERT DATA {
  GRAPH <http://mu.semte.ch/graphs/administrative-unit> {
    <http://data.lblod.info/id/bestuurseenheden/7cab1a9c6f5dc798164f1c6812fc790e5279a8fe7800a2e8b7a159c581c866f1>
      skos:prefLabel "Stadsregio Turnhout" ;
      <http://www.w3.org/ns/org#hasPrimarySite> <http://data.lblod.info/id/vestigingen/e215e097cd2f054c9de6db80cf34d65d> .
  }
}

restart cache & resource

Then visit http://localhost:4200/bestuurseenheden/7cab1a9c6f5dc798164f1c6812fc790e5279a8fe7800a2e8b7a159c581c866f1/leidinggevenden !


To import leidinggevenden:
- dr inspect app-organization-portal-leidinggevenden-consumer-1
- curl --request POST <ip>/flush // Wait that the flushing finishes
- drc restart leidinggevenden-consumer // Wait for the initial sync to finish, +/-10min
- sh ./scripts/reset-elastic.sh // This takes +/- 1h, check search logs
```

### DIFF
--- a/config/migrations/2023/20230712152618-igs-link-positions-to-unit-classifications.sparql
+++ b/config/migrations/2023/20230712152618-igs-link-positions-to-unit-classifications.sparql
@@ -1,0 +1,12 @@
+PREFIX ext: <http://mu.semte.ch/vocabularies/ext/>
+
+INSERT DATA {
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    <http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/41caf7e6-b040-4720-9cc2-a96cfffed5b4>
+      ext:appliesWithin
+        <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/b156b67f-c5f4-4584-9b30-4c090be02fdc> ,
+        <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/d01bb1f6-2439-4e33-9c25-1fc295de2e71> ,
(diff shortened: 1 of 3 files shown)


bdevloed:
Postal codes seem to be missing looks like a bug previously in leidinggevenden - discovered it last week
```
SELECT DISTINCT ?p {
GRAPH <http://mu.semte.ch/graphs/landing-zone/leidinggevenden> {
  ?s
    a <http://www.w3.org/ns/locn#Address>;
    ?p ?o
  }
}
```

claire-lovisa:
@bdevloed Thanks for the catch ! I sent a PR fo review here https://github.com/lblod/app-digitaal-loket/pull/435 to fix this postal codes issue.

As it's not related to this PR, I think it's safe to merge if the code is OK for you, what do you think ?

bdevloed:
@claire-lovisa Indeed. Thanks for creating the other PR at loket as well! :)

bdevloed (approved):

PR #302, opened 2023-07-13
Include local-identifier and label for each identifier of organizations sent to ldes feed

elpoelma:
### Overview
This PR includes a change to the `dispatch` function of the ldes-delta-pusher. For each organization sent to the LDES feed, the local identifiers and labels for each of their identifiers is also included.

### How to test
- Set-up the stack
- Create/edit and organization and include some identifiers
- Check if the ldes-feed has been updated with the new information on `/ldes/organizations/1`, the new information will be included on the last page of the feed.

### Notes
The structure sent to the LDES-feed is a bit complicated, maybe we can use other predicates to link the KBO/OVO numbers to the organizations directly instead of through the `local-identifier`/`structured-identifier` predicates?


elpoelma:
Note: when we deploy the new version on dev, we need to make sure to clear the LDES feed in order for the data on it to be complete and consistent.

elpoelma:
I'm gonna merge this for now, we can always add changes to the data-model later.

PR #307, opened 2023-07-20
Import OVO codes for districts, worship and central worship services

claire-lovisa:
OP-2531

To test this PR, you can run the migrations and then via the frontend open districts and worship services to check if the OVO codes are showing as expected.


lagartoverde (approved):
looks good to me

PR #316, opened 2023-08-16
add migration to import igs from csv

lagartoverde:
OP-2528 and OP-2529


claire-lovisa (changes requested):
Hello ! I have some change requests, the current migration misses some data and adds erronated uuids. Here are the improvments I see:
1. Move the migrations in the folder `20230704112900-onboarding-igs`
2. It's a ttl so probably no need to split it in two migrations ?
3. Many KBO have already been imported in `20230704112912-import-igs-from-loket.ttl` (all the KBOs of organizations that existed previously in Loket), causing double KBOs as a new URI has been added for them in this PR
4. For the new IGS that were not existing in Loket already, we also need to generate bestuursorganen, bestuursorganen in time and bestuursfuncties
5. I don't see addresses in the migration (the resources are there but not the content of the addresses), is it expected ? If so, we'll probably have to add them later on
6. Some uuids are added twice, which would be OK if the values were the same, but the values of the uuids of besturen are different because of an extra ">" at the end:
   ```
    <http://data.lblod.info/id/bestuurseenheden/c9fd02221ff5b1f0e671b6a155bef39f9aa6232e351eb4f0f00239f677d51643> a besluit:Bestuurseenheid, org:Organization;
      mu:uuid "c9fd02221ff5b1f0e671b6a155bef39f9aa6232e351eb4f0f00239f677d51643>";
   ```

claire-lovisa:
For the missing organen, organen in time and functies, you can take inspiration on https://github.com/lblod/app-organization-portal/blob/development/config/migrations/2023/20230320135902-onboarding-agb-apb/20230801140025-add-missing-organen-and-functies-to-agbs-and-apbs.sparql which does a similar job for agbs and apbs. I think that the only difference is that IGS have one more orgaan of classification `http://data.vlaanderen.be/id/concept/BestuursorgaanClassificatieCode/5733254e-73ff-4844-8d43-7afb7ec726e8`.

claire-lovisa (changes requested):
Hey ! It's much better with this update :smile: the indexing goes well, I can browse IGS. I just still noticed a small quirk in the data:
- We still have double KBOs for besturen already existing in Loket, such as <http://data.lblod.info/id/bestuurseenheden/efe77fae-7ba7-4a62-873f-1f291b5a911f>. <http://data.lblod.info/id/identificatoren/4de90fb6-1be3-4724-9e32-6de678fe6a7a> and <http://data.lblod.info/id/identificatoren/7a95a7f875004dd5da1cf5407ab9a298> are both representing the same KBO
- New besturen, such as <http://data.lblod.info/id/bestuurseenheden/95f24dbe-c4c1-4735-8d76-ec296e603295>, don't have a KBO (maybe it's expected ..?)

lagartoverde:
> Hey ! It's much better with this update 😄 the indexing goes well, I can browse IGS. I just still noticed a small quirk in the data:
> 
> * We still have double KBOs for besturen already existing in Loket, such as http://data.lblod.info/id/bestuurseenheden/efe77fae-7ba7-4a62-873f-1f291b5a911f. http://data.lblod.info/id/identificatoren/4de90fb6-1be3-4724-9e32-6de678fe6a7a and http://data.lblod.info/id/identificatoren/7a95a7f875004dd5da1cf5407ab9a298 are both representing the same KBO
> * New besturen, such as http://data.lblod.info/id/bestuurseenheden/95f24dbe-c4c1-4735-8d76-ec296e603295, don't have a KBO (maybe it's expected ..?)

My fault, I did the contrary to what I had to do xd, I removed the kbos from the ones not in loket, it should be good now :)

claire-lovisa (approved):
Looks good ! When you deploy to dev, can you also reindex (`sh ./scripts/reset-elastic`) so that the newly added IGS are displayed in the main page ? Thx :)

PR #317, opened 2023-08-23
Generate functies for new agbs and apbs

claire-lovisa:


lagartoverde (approved):
I mean I copied your code to my migration so it should be correct or we are both wrong 🙃

PR #319, opened 2023-08-29
Add AGBs and APBs related organizations

claire-lovisa:
OP-2591

Note: this migration predates on purpose the migration updating the related organizations model on branch `feature/memberships`, to be able to review/deploy faster.


lagartoverde (approved):
Looks good :)

PR #331, opened 2023-09-18
Add report generation service + orgs report

claire-lovisa:
OP-2711 & OP-2714

The goal, for now, is to provide multiple endpoints with different ways of retrieving the reports (in bulk in an archive and one by one as files). They'll be used by the team that required those reports to test their flow and see what way of retrieving the reports works best for them. The available endpoints might change in the future depending on their feedback.

For now only one report has been configured so that we can have some data to test. In the future, we'll add more of them.

**How to manually generate a report** _(otherwise they run via cron jobs)_

```
dr inspect app-organization-portal-report-generation-1

curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"data":{"attributes":{"reportName":"organizations"}}}' \
  <ip-report-generation-service>/reports
```

**How to call the endpoints**

```
dr inspect app-organization-portal-download-exports-1

curl <ip-download-exports-service>/export-names
curl <ip-download-exports-service>/exports
curl <ip-download-exports-service>/exports/:exportName
```


mirdono (approved):

config/reports/organizations-report.js
```diff
@@ -0,0 +1,116 @@
+import { generateReportFromData, batchedQuery } from '../helpers.js';
+import { querySudo as query } from '@lblod/mu-auth-sudo';
```
aliokan, line 2: `query` is never used ;)

PR #336, opened 2023-09-27
bump mu-search to 0.9.0

nvdk:
pretty much a drop in upgrade
https://github.com/mu-semtech/mu-search/blob/master/CHANGELOG.md#v090


nbittich:
It was already bumped, thanks 👍🏼

PR #350, opened 2023-11-20
Refactor: Introduce generic Agent resource

mirdono:
OP-2652

Corresponding [frontend PR](https://github.com/lblod/frontend-organization-portal/pull/541)

Note: requires re-indexing search to work properly


claire-lovisa (changes requested):
Small change requests, in the meantime I'm running a reindex to verify that everything behaves as expected after a refresh of ES :)

claire-lovisa (approved):

PR #353, opened 2023-11-25
Fix dev mandatarissen without persons

claire-lovisa:
OP-2868

Allows dev to display its data properly before we do a proper fix of the consumer

This migration needs to be followed by a reindex

### DIFF
--- a/config/dev-migrations/20231104231300-remove-mandatarissen-without-person.sparql
+++ b/config/dev-migrations/20231104231300-remove-mandatarissen-without-person.sparql
@@ -1,4 +1,5 @@
 PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
+PREFIX org: <http://www.w3.org/ns/org#>
 
 DELETE {
   GRAPH ?g {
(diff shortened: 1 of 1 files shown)


mirdono:
The inclusion of the `?mandaat a ?type` triple in the `WHERE` clause causes this migrations to miss a number of imo relevant entries. The following query returns 8717 results after executing the migration on a DB that was freshly synced with remote DEV.
```
PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
SELECT (COUNT (DISTINCT ?mandaat) AS ?count)
  FROM <http://mu.semte.ch/graphs/administrative-unit>
  FROM <http://mu.semte.ch/graphs/worship-service>
  WHERE {
    ?mandaat mandaat:isBestuurlijkeAliasVan ?person .
    FILTER NOT EXISTS { ?person ?p ?o . }
  }
```

mirdono (changes requested):
The inclusion of the `?mandaat a ?type` triple in the WHERE clause causes this migrations to miss a number of imo relevant entries. The following query returns 8717 results after executing the migration on a DB that was freshly synced with remote DEV.
```
PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#>
SELECT (COUNT (DISTINCT ?mandaat) AS ?count)
  FROM <http://mu.semte.ch/graphs/administrative-unit>
  FROM <http://mu.semte.ch/graphs/worship-service>
  WHERE {
    ?mandaat mandaat:isBestuurlijkeAliasVan ?person .
    FILTER NOT EXISTS { ?person ?p ?o . }
  }
```

claire-lovisa:
@mirdono I added a second step to the migration that removes the mandatarissen without a type !

mirdono (approved):

PR #355, opened 2023-12-05
converted-date-to-datetime

Dachamelemon:
OP-2557

### DIFF
--- a/config/resources/domain.json
+++ b/config/resources/domain.json
@@ -92,11 +92,11 @@
       "super": ["agents"],
       "attributes": {
         "agent-start-date": {
-          "type": "date",
+          "type": "datetime",
           "predicate": "ch:startdatum"
         },
         "agent-end-date": {
-          "type": "date",
+          "type": "datetime",
           "predicate": "ch:eindedatum"
         }
       },
@@ -140,11 +140,11 @@
       "super": ["agents-in-position"],
       "attributes": {
         "start-date": {
-          "type": "date",
+          "type": "datetime",
           "predicate": "mandaat:start"
         },
         "end-date": {
-          "type": "date",
+          "type": "datetime",
           "predicate": "mandaat:einde"
         }
       },
@@ -188,7 +188,7 @@
       "super": ["mandatories"],
       "attributes": {
         "expected-end-date": {
-          "type": "date",
+          "type": "datetime",
           "predicate": "ere:geplandeEinddatumAanstelling"
         },
         "reason-stopped": {
--- a/config/migrations/2023/20231205094320-mandaat-date-to-datetime.sparql
+++ b/config/migrations/2023/20231205094320-mandaat-date-to-datetime.sparql
@@ -1,0 +1,110 @@
+PREFIX mandaat: <http://data.vlaanderen.be/ns/mandaat#> 
+PREFIX ch: <http://data.lblod.info/vocabularies/contacthub/>
+PREFIX ere: <http://data.lblod.info/vocabularies/erediensten/>
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+DELETE {
+  GRAPH ?g {
+  ?s mandaat:start ?o .
+  }
+}
+INSERT {
+ GRAPH ?g {
+   ?s mandaat:start ?converteddate .
+ }
+}
+WHERE {
+  GRAPH ?g {
+    ?s mandaat:start ?o .
+  }
+  VALUES ?g {
+    <http://mu.semte.ch/graphs/worship-service>
+    <http://mu.semte.ch/graphs/administrative-unit>
+ }
+ BIND(xsd:dateTime(?o) AS ?converteddate).
+}
+;
+DELETE {
+    GRAPH ?g {
+    ?s mandaat:einde ?o .
+    }
+  }
+  INSERT {
+   GRAPH ?g {
+     ?s mandaat:einde ?converteddate .
+   }
+  }
+  WHERE {
+    GRAPH ?g {
+      ?s mandaat:einde ?o .
+    }
+    VALUES ?g {
+      <http://mu.semte.ch/graphs/worship-service>
+      <http://mu.semte.ch/graphs/administrative-unit>
+   }
+   BIND(xsd:dateTime(?o) AS ?converteddate).
+  }
+
+;
+DELETE {
+  GRAPH ?g {
+  ?s ch:startdatum ?o .
+  }
+}
+INSERT {
+ GRAPH ?g {
+   ?s ch:startdatum ?converteddate .
+ }
+}
+WHERE {
+  GRAPH ?g {
+    ?s ch:startdatum ?o .
+  }
+  VALUES ?g {
+    <http://mu.semte.ch/graphs/worship-service>
+    <http://mu.semte.ch/graphs/administrative-unit>
+ }
+ BIND(xsd:dateTime(?o) AS ?converteddate).
+}
+;
+DELETE {
+    GRAPH ?g {
+    ?s ch:eindedatum ?o .
+    }
+  }
+  INSERT {
+   GRAPH ?g {
+     ?s ch:eindedatum ?converteddate .
+   }
+  }
+  WHERE {
+    GRAPH ?g {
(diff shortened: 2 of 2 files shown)


mirdono:
If I understand the [reasoner rules](https://github.com/lblod/app-organization-portal/blob/54e381bfad96b373475ba724b40d5d4e4c5f9452/config/reasoner/dl2op/main/mapping.n3#L124) concerning dates correctly the `datetime` literals received from Loket are currently converted to `date` literals. These rules should probably be changed (or removed) to avoid unnecessary modifications of incoming `datetime` literals.

config/resources/domain.json
```diff
@@ -92,11 +92,11 @@
       "super": ["agents"],
       "attributes": {
         "agent-start-date": {
-          "type": "date",
+          "type": "datetime",
           "predicate": "ch:startdatum"
```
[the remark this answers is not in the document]
claire-lovisa: @Dachamelemon This is still missing from the current configuration from what I can see ? Could you add the date to datetime rule for them in the reasoner ? Thanks !

mirdono:
@Dachamelemon could you verify whether the migration works as intended for you? On my end it does not seem to have any effect on the data, i.e. none of the dates are actually converted to datetimes. I have experimented a bit with the delete/insert queries from your migration, e.g. applying it to one specific triple, but it never seems to have any effect (or results in an error).

config/reasoner/dl2op/main/mapping.n3
```diff
@@ -121,6 +107,46 @@ PREFIX lblodlg: <http://data.lblod.info/vocabularies/leidinggevenden/>
...
+    ?s ch:eindedatum ?startDate
+  } e:becomes {
+    ?s ch:eindedatum ?startDateTime
+  }.
+} => {
+  ?s ch:eindedatum ?startDateTime
+}.
+
+{
+  ?s ere:geplandeEinddatumAanstelling ?startDDate.
```
mirdono, line 137: The `ere` prefix is not defined in this file, causing the consumers to crash:
```
leidinggevenden-consumer-1  | Statements:  ["{\"errors\":[{\"detail\":\"** ERROR ** file:///config/dl2op/main/mapping.n3 ** no_prefix_directive(ere,after_line(134))\\n\",\"status\":500}]}",""]
```

There is also a **D** too much in the object name.

mirdono (changes requested):

config/migrations/2023/20231205094323-mandaat-date-to-datetime.sparql
mirdono, line 1: The migration is lacking rules for `mandaat:bindingStart` and `mandaat:bindingEinde`.  At least the DEV environment contains date entries for these predicates.

config/reasoner/dl2op/main-info/mapping.n3
```diff
@@ -68,34 +68,84 @@ PREFIX ere: <http://data.lblod.info/vocabularies/erediensten/>
...
+    ?s ch:startdatum ?startDate
+  } e:becomes {
+    ?s ch:startdatum ?startDateTime
+  }.
+} => {
+  ?s ch:startdatum ?startDateTime
+}.
+
+{
+  ?s ch:eindedatum ?startDDate.
```
mirdono, line 125: The object variable contains two **D**s.

config/reasoner/dl2op/main/mapping.n3
```diff
@@ -67,25 +67,11 @@ PREFIX lblodlg: <http://data.lblod.info/vocabularies/leidinggevenden/>
...
-  ?s mandaat:start ?startDate
-}.
-
 # ex:s <http://data.vlaanderen.be/ns/mandaat#einde> "2019-02-01T10:32:52+00:00"^^xsd:dateTime.
 {
   ?s mandaat:einde ?eindeDateTime.
   # use backward arrow rule defined below
-  ?eindeDateTime :dateFromDateTime ?eindeDate.
+  ?eindeDateTime :dateTimeFromDate ?eindeDate.
   {
```
mirdono, line 75: Loket already uses `dateTime` for `mandaat:einde` (and `mandaat:start`), so this conversion is not necessary. This remark also applies to the same rule in `main-info/mapping.n3`.

mirdono (changes requested):

config/migrations/2023/20231205094323-mandaat-date-to-datetime.sparql
mirdono: @Dachamelemon Could you process the above comment as well? Furthermore, can you also rename the migrations file when doing so? Otherwise it will not be executed again.

mirdono (changes requested):
In Loket the `mandaat:bindingStart` and `mandaat:bindingEinde` are of type `date` (cf. [configuration](https://github.com/lblod/app-digitaal-loket/blob/cb4026d044cb399b96bc8d539ec5ec7bbebe393a/config/resources/slave-besluit-domain.lisp#L104)).

Assuming this PR should cover these predicates:
- appropriate queries should be added to the migration (as mentioned before); and
- the rules in the mapping files should use the new function `dateTimeFromDate` to convert the incoming dates instead of the old `dateFromDateTime`.

config/migrations/2023/20231205094325-mandaat-date-to-datetime.sparql
mirdono, line 1: I am also still having problems with getting consistent results from this migration. Its execution does not seem to have immediate effect, yet after some random amount of time/stack restarts the conversions appear to have occurred. I suggest to check with someone more knowledgeable on SPARQL/virtuoso concerning this matter to ensure that the current method is the correct (and/or best) one.

Note, looking at similar migrations, e.g. [here](https://github.com/lblod/app-organization-portal/blob/99f5f5365bc3b5a958d9920f403eb55bc7167e82/config/migrations/2023/20230503143253-organen-start-and-end-should-be-datetimes-part-one.sparql#L1) or [here](https://github.com/lblod/app-organization-portal/blob/99f5f5365bc3b5a958d9920f403eb55bc7167e82/config/migrations/2023/20230320135902-onboarding-agb-apb/20230801181808-convert-datetime-to-date-agb-apb-organen-in-time.sparql#L22), I did notice they tend to use a different method for conversion. The latter example also contains a comment about a possibly relevant bug in virtuoso.

mirdono (changes requested):

mirdono (changes requested):
There are discrepancies between the datetime elements resulting from the migration and those imported by the consumers. The datetime elements imported by a consumer contain a timezone `Z` as affix whereas the results of the migration do not. Furthermore, the migration results have as time `02:00:00` (to work around the virtuoso issue) while the results from the consumers have `00:00:00Z` as time. This results in the datetime elements synchronized from Loket being inserted next to any already existing elements instead of overwriting them, in turn leading to wrong/outdated dates being shown in the frontend.

I did some quick tests as to how to deal with this issue. I found that modifying the migration to append a 'Z' in the bindings and setting the time back to 00:00:00 seems to work as expected, without breaking the migration again due to the virtuoso thing. I did not test this thoroughly so I may be missing something. I do think that adding the 'Z' changes the semantics of the datetime as it implies it should be interpreted as a UTC date, which is technically not correct for this application. (Recall also the timezone magic in the original `dateFromDateTime` rule.) It is probably a good idea to check whether this has any side effects now or for the future.

As a side note, I first tried changing the time concatenated in the reasoner rules but this did not seem to have any impact on the resulting elements, In other words, the imported datetime elements kept having `00:00:00Z` as time. (I also do not immediately see where the 'Z' actually comes from.) This makes me wonder whether the `dateTimeFromDate` rule actually works as intended. Unfortunately my knowledge here is too limited to assess them in depth.

mirdono:
I noticed that the mapping in [private-info](https://github.com/lblod/app-organization-portal/blob/development/config/reasoner/dl2op/private-info/mapping.n3) also applies the old `dateFromTime` rule for `mandaat:start` and `mandaat:einde`. This should most likely be adapted in the same fashion as the other mappings to avoid we end up with inconsistent data.

mirdono (approved):
Looks good to me

PR #356, opened 2023-12-06
Add publication triplestore and organizations public info producer

claire-lovisa:
OP-2875

In this PR, we add a new producer called `organizations-public-info`. It is basically a copy of the `public` producer, without the restrictions to only produce worship organizations data. This one produces a wider range of data that will first be used by the CLBV team for their applications and then, in a second phase, will replace the `public` producer in Loket as well.

We also introduce a new triplestore to the application, the `publication-triplestore`. This is inspired by Loket and the issues they had there when they started to have multiple producers graph in the main triplestore, slowing down a lot the main triplestore and consequently the application. To avoid getting the same issues in OP, let's try to anticipate!


lagartoverde (approved):
Looks good, but now that we are doing this 2 dbs thing, shouldn't we move everything to the secondary db?

PR #359, opened 2023-12-08
Fix: delete unnecessary change event

mirdono:
OP-2879

The migration needs to be tested against PROD data.

### DIFF
--- a/config/migrations/2023/20231208160845-delete-change-event-lievegem.sparql
+++ b/config/migrations/2023/20231208160845-delete-change-event-lievegem.sparql
@@ -1,15 +1,21 @@
 PREFIX core: <http://mu.semte.ch/vocabularies/core/>
 PREFIX org: <http://www.w3.org/ns/org#>
 PREFIX lblodorg: <http://lblod.data.gift/vocabularies/organisatie/>
+PREFIX euro: <http://data.europa.eu/m8g/>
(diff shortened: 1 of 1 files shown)


aliokan (commented):
@mirdono I havn't write a lot of DELETE migrations, but I'm curious to know what will be deleted here. 

Is `DELETE WHERE` equivalent to ` DELETE * WHERE` ?

claire-lovisa (changes requested):
The change event also has a decision (`http://data.europa.eu/m8g/hasFormalFramework`) that needs to be removed I think ?

mirdono:
> @mirdono I havn't write a lot of DELETE migrations, but I'm curious to know what will be deleted here.
> 
> Is `DELETE WHERE` equivalent to ` DELETE * WHERE` ?

I am not a SPARQL expert either, but from what I understand from the [specification](https://www.w3.org/TR/sparql11-update/#deleteInsert) that seems to be correct. Specifically the specification's paragraph about the `GroupGraphPattern` in the `WHERE` clause:

> _"The [GroupGraphPattern](http://www.w3.org/TR/sparql11-query/#rGroupGraphPattern) in the WHERE clause is evaluated as in a SPARQL query "SELECT * WHERE GroupGraphPattern" and all the solution bindings are applied to the preceding DELETE and INSERT templates for defining the triples to be deleted from or inserted into the Graph Store."_

claire-lovisa (approved):

PR #362, opened 2023-12-14
Ignore publication triplestore data folder

claire-lovisa:
To avoid thinking we have untracked changes when it's just the new publication triplestore :grin:


Dachamelemon (commented):
Perfect!

mirdono (approved):

PR #366, opened 2023-12-21
Bugt/fix duplicate lang strings

lagartoverde:
https://binnenland.atlassian.net/browse/OP-2561

### DIFF
--- a/config/migrations/2023/20231221132600-fix-duplicate-lang-strings.sparql
+++ b/config/migrations/2023/20231221132600-fix-duplicate-lang-strings.sparql
@@ -7,7 +7,7 @@ DELETE
 WHERE {
   GRAPH ?g {
     ?s ?p ?oLang.
-    ?a ?b ?o .
+    ?a ?p ?o .
   }
 
   BIND(LANG(?oLang) as ?lang)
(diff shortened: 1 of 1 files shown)


config/migrations/2023/20231221132600-fix-duplicate-lang-strings.sparql
```diff
@@ -0,0 +1,16 @@
+DELETE
+{
+  GRAPH ?g {
+    ?s ?p ?oLang.
+  }
+}
+WHERE {
+  GRAPH ?g {
+    ?s ?p ?oLang.
+    ?a ?b ?o .
```
bdevloed, line 10: Shouldn't the subject and property match as well?

i.e.:
```
?s ?p
  ?oLang,
  ?o.
```

bdevloed (changes requested):
I think the current migration might delete literals belonging to different subjects which happen to have the same literal value.

config/migrations/2023/20231221132600-fix-duplicate-lang-strings.sparql
lagartoverde: You are right, I based it on a query made by Claire and didn't realise that problem, I fixed it now

bdevloed (approved):

PR #367, opened 2023-12-21
Feature: onboarding public ocmw associations

mirdono:
OP-2841, OP-2843, OP-2844

Summary:
- Added data migrations for new classification codes for the different OCMW associations, currently only the two for public OCMW associations (Welzijnsvereniging and Autonome verzorgingsinstelling) are used.
- Reasoner: added rule to sync functionaries with Loket
- Resource: generalised the `was-founded-by-organization` to a many cardinality as OCMW associations can have multiple founders

Frontend PR: https://github.com/lblod/frontend-organization-portal/pull/553


lagartoverde (approved):
Looks good to me, but maybe wait for another reviewer who has more context about the data.
One small remark is that you changed a lot of styling things in the config.ex so it's harder to see your changes, maybe the next time, I would make a PR only to fix these kind of things

PR #368, opened 2024-01-09
Feature: Added new classification codes for PEVAs

mirdono:
Adds new classification codes for PEVA municipality and PEVA province which will be onboarded in OP-2640. These new classification codes are specifically defined in the context of OP and are not (yet) part of the upstream standard. 

Frontend PR: https://github.com/lblod/frontend-organization-portal/pull/558


aliokan (approved):
It looks good, I haven't test it

PR #370, opened 2024-01-17
Feat/replace site type

lagartoverde:
OP-2906 linked to https://github.com/lblod/frontend-organization-portal/pull/563


lagartoverde:
> Changing all the graph currently would apply changes to the graphs for the consumers and producers and this could cause issues overtime.
> 
> Graphs that could be changed would be the "administrative-unit" and "worship-services" graphs. there is also "http://mu.semte.ch/graphs/public" (because here there is some overlapping data between OP and WOP).

In this case only the administrative-unit graph is relevant, changed it

Dachamelemon (approved):
Looks good!

PR #372, opened 2024-01-31
Fix: combine multiple incorrect merger change events into a single one change event

mirdono:
OP-2973

Note: this migration should be tested against PROD data


nbittich:
lgtm! didn't try it tho, @bdevloed could check it as well to make sure nothing is missing, but seems ok to me

mirdono:
Superseded by #377 which allows to release a hotfix containing the same migration as introduced here.

PR #373, opened 2024-01-31
Feat/add nis to producer

lagartoverde:


config/migrations/2024/20240130104000-add-export-concept-to-nis-concepts.sparql
```diff
@@ -0,0 +1,14 @@
...
+
+
+
+INSERT  {
+   GRAPH <http://mu.semte.ch/graphs/public> {
+     ?concept skos:inScheme <http://lblod.data.gift/concept-schemes/0d101f02-609d-4cde-bb07-639d874470c4>.
+   }
+}
+WHERE {
+    ?adminUnit besluit:werkingsgebied ?location.
```
bdevloed, line 12: Maybe safer to use this as a where, now it's only by chance that no other geo schemes match (because of they contain language tags in the labels):

```?concept <http://www.w3.org/2004/02/skos/core#inScheme> <http://vocab.belgif.be/auth/refnis2019>.```

However, it seems to do the job.

bdevloed (approved):

config/migrations/2024/20240130104000-add-export-concept-to-nis-concepts.sparql
lagartoverde: Mmmm, I want to export all concepts that are related to the locations, so I think limiting just to that scheme is not something I would want to do, right?

bdevloed: Ok, nvm the remark .

PR #332, opened 2023-09-20
Fix/boolean

bdevloed:
configure mu-cl-resources to use xsd:booleans and migrate custom booleans to xsd

PR #374, opened 2024-02-02
Fix/resource heavy migrations

mirdono:
# Context
Two recently added migrations are rather resource intensive to execute and can cause problems for virtuoso, which in turn causes the migrations service to crash. The migrations in question are those introduced in PRs [boolean type fix](https://github.com/lblod/app-organization-portal/pull/332) and [removal of duplicate language strings](https://github.com/lblod/app-organization-portal/pull/366).
At the time of writing neither of these migrations have been executed in our remote DEV environment so it is unclear whether that environment would be able to handle them properly. In my local setup these migrations to reliably crash the migrations service, irrelevant of whether I have loaded DEV or PROD data in the triplestore.

The full error with which the migrations service crashes is the following:
```
/usr/local/lib/ruby/2.5.0/net/protocol.rb:181:in `rbuf_fill': too many connection resets (due to Net::ReadTimeout - Net::ReadTimeout) after 0 requests on 50201713406840, last used 1706696178.4719708 seconds ago (Net::HTTP::Persistent::Error)
 	from /usr/local/lib/ruby/2.5.0/net/protocol.rb:157:in `readuntil'
 	from /usr/local/lib/ruby/2.5.0/net/protocol.rb:167:in `readline'
 	from /usr/local/lib/ruby/2.5.0/net/http/response.rb:40:in `read_status_line'
 	from /usr/local/lib/ruby/2.5.0/net/http/response.rb:29:in `read_new'
 	from /usr/local/lib/ruby/2.5.0/net/http.rb:1494:in `block in transport_request'
 	from /usr/local/lib/ruby/2.5.0/net/http.rb:1491:in `catch'
 	from /usr/local/lib/ruby/2.5.0/net/http.rb:1491:in `transport_request'
 	from /usr/local/lib/ruby/2.5.0/net/http.rb:1464:in `request'
 	from /usr/local/bundle/gems/net-http-persistent-3.1.0/lib/net/http/persistent.rb:964:in `block in request'
 	from /usr/local/bundle/gems/net-http-persistent-3.1.0/lib/net/http/persistent.rb:662:in `connection_for'
 	from /usr/local/bundle/gems/net-http-persistent-3.1.0/lib/net/http/persistent.rb:958:in `request'
 	from /usr/local/bundle/gems/sparql-client-3.0.1/lib/sparql/client.rb:696:in `request'
 	from /usr/local/bundle/gems/sparql-client-3.0.1/lib/sparql/client.rb:344:in `response'
 	from /usr/local/bundle/gems/sparql-client-3.0.1/lib/sparql/client.rb:305:in `query'
 	from /usr/local/bundle/gems/mu-auth-sudo-0.1.0/lib/mu/auth-sudo.rb:22:in `query'
 	from /usr/src/app/ext/web.rb:209:in `query_sudo_with_logging'
 	from /usr/src/app/ext/web.rb:66:in `execute!'
 	from /usr/src/app/ext/web.rb:151:in `block in execute_migrations'
 	from /usr/src/app/ext/web.rb:149:in `each'
 	from /usr/src/app/ext/web.rb:149:in `execute_migrations'
 	from /usr/src/app/ext/web.rb:214:in `boot'
 	from /usr/src/app/ext/web.rb:217:in `<top (required)>'
 	from web.rb:87:in `require_relative'
 	from web.rb:87:in `<main>'
```

# Proposed solution
I rewrote both migrations by
- Explicitly limiting them both to our 3 primary graphs `administrative-unit`, `worship-service`, and `public`. Additionally, this prevents changes in the graphs used by producers and consumers (e.g. the `landing-zone` graphs), which I have been advised we should not edit other than via the services using these graphs.
- For the duplicate lang strings migration, I also moved the second triple in the WHERE-clause behind the first FILTER-statement to avoid matches that would be filtered out anyway by that statement.

These changes seem to solve the migrations service crash at least on my system,  using DEV as well as PROD data. For completeness, I tested the results of the lang string migration with the following query derived from the original migration (warning takes a while to execute):

```sparql
SELECT ?g ?oLang ?o {
   GRAPH ?g {
    ?s ?p ?oLang.
    ?s ?p ?o .
  }

  BIND(LANG(?oLang) as ?lang)
  FILTER(STRLEN(STR(?lang)) > 0)
  FILTER(STR(?oLang) = ?o)


  VALUES ?g {
    <http://mu.semte.ch/graphs/public>
    <http://mu.semte.ch/graphs/worship-service>
  }
}
```


aliokan (approved):
looks good ! :+1:

PR #382, opened 2024-02-22
Fix: remove already imported founders for public OCMW associations

mirdono:
The data concerning related organisations (OP-2979) for onboarding the public OCMW associations is not yet finalised by business. In order to prevent this from blocking the next (pre-)release this PR removes the relevant, already imported draft data from the DEV environment.

What I did:
- Removed the original migration such that it is not executed on QA (and PROD).
- Added a DEV-specific migration that removes the triples added by the original migration


aliokan (approved):
Looks good :+1:

PR #383, opened 2024-02-22
Feat/clb consumer

lagartoverde:
The sync is working, so the code can be reviewed
BUT: CLB is doing weird things with the edit mode that I have to check so the data that gets edited and the data that gets saved to the database might differ and also the telephone format is different so I will discuss that with Sofie
DO NOT MERGE


config/clb-consumer/contact-data/config.js
```diff
@@ -0,0 +1,32 @@
...
+const BYPASS_MU_AUTH_FOR_EXPENSIVE_QUERIES = process.env.BYPASS_MU_AUTH_FOR_EXPENSIVE_QUERIES == 'true' ? true : false;
+const DIRECT_DATABASE_ENDPOINT = process.env.DIRECT_DATABASE_ENDPOINT || 'http://virtuoso:8890/sparql';
+const MAX_DB_RETRY_ATTEMPTS = parseInt(process.env.MAX_DB_RETRY_ATTEMPTS || 5);
+const MAX_REASONING_RETRY_ATTEMPTS = parseInt(process.env.MAX_REASONING_RETRY_ATTEMPTS || 5);
+const SLEEP_BETWEEN_BATCHES = parseInt(process.env.SLEEP_BETWEEN_BATCHES || 100);
+const SLEEP_TIME_AFTER_FAILED_REASONING_OPERATION = parseInt(process.env.SLEEP_TIME_AFTER_FAILED_REASONING_OPERATION || 10000);
+const SLEEP_TIME_AFTER_FAILED_DB_OPERATION = parseInt(process.env.SLEEP_TIME_AFTER_FAILED_DB_OPERATION || 60000);
+
+// Graphs
+const TARGET_GRAPH = process.env.TARGET_GRAPH || `http://mu.semte.ch/graphs/organizations`;
```
mirdono, line 12: **Suggestion:** Although the value of `TARGET_GRAPH` currently does not seem to be used, I think it would be better to default to a graph we use in OP. For example, `http://mu.semte.ch/graphs/administrative-unit` as is done for some other consumers.

config/clb-consumer/contact-data/config.js
```diff
@@ -0,0 +1,32 @@
...
+const LANDING_ZONE_GRAPH = process.env.DCR_LANDING_ZONE_GRAPH || `http://mu.semte.ch/graphs/landing-zone/op`;
```
mirdono, line 13: **Nitpick:** I would use a more meaningful name for the fallback value here. Currently, should the docker-compose configuration be (accidentally) changed in the future to no longer set the `DCR_LANDING_ZONE_GRAPH` we would  end up with a `http://mu.semte.ch/graphs/landing-zone/op` graph. That might cause some confusion as to what the purpose of such a graph is.

config/clb-consumer/contact-data/delta-sync-dispatching.js
```diff
@@ -0,0 +1,123 @@
...
+*              predicate: "<http://baz>",
+*              object: "<http://boom>^^<http://datatype>"
+*            }
+*         ]
+* @return {void} Nothing
+*/
+
+const publicTypes = [
+  '<http://lblod.data.gift/vocabularies/organisatie/TypeVestiging>'
+]
```
mirdono, line 29: **Nitpick:** Please do not add code in between a function and its documentation.

config/clb-consumer/contact-data/delta-context-config.js
```diff
@@ -0,0 +1,256 @@
...
+          {
+            ?site org:siteAddress ?contact.
+            ?contact ?pContact ?oContact.
+          }
+          UNION
+          {
+            ?site organisatie:bestaatUit ?address.
+            ?address ?pAddress ?oAddress.
+          }        
+          BIND(IRI(CONCAT("http://mu.semte.ch/graphs/organizations/", ?uuid)) AS ?graph)
```
mirdono, line 156: **Question:** Is this line necessary? The `?graph` variable is not used as far as I see. (In case the variable is meant to be used the graph URI should be corrected.)

config/clb-consumer/contact-data/delta-context-config.js
```diff
@@ -0,0 +1,256 @@
...
```
mirdono, line 203: Same as above.

config/clb-consumer/contact-data/delta-sync-dispatching.js
```diff
@@ -0,0 +1,123 @@
...
+async function dispatch(lib, data) {
+  const { mu, fetch } = lib;
```
mirdono, line 31: **Polish:**: Neither `mu` or `fetch` are actually used ion the following code.

config/clb-consumer/contact-data/delta-sync-dispatching.js
```diff
@@ -0,0 +1,123 @@
...
+      const typeTriple = contextTriples.find((context) => context.subject === subject && context.predicate === '<http://www.w3.org/1999/02/22-rdf-syntax-ns#type>');
+      if(!typeTriple) continue;
+      const type = typeTriple.object;
+      if(publicTypes.includes(type)) {
+        deletesOnPublic.push(`${deletion.subject} ${deletion.predicate} ${deletion.object}.`)
+        for(let triple of otherContextTriples) {
+          deletesOnPublic.push(`${triple.subject} ${triple.predicate} ${triple.object}.`)
+        }
+      }
+    }
```
mirdono, line 111: **Suggestion:** While the above seems functionally correct it is rather hard to parse/read imo. Would you mind adding some formatting such as blank lines and break up some long lines? Furthermore, several (if-then-else) blocks are nearly the same. Imo it could make the code easier to understand/maintain if we split these common parts into reusable functions.

config/clb-consumer/contact-data/initial-sync-dispatching.js
```diff
@@ -0,0 +1,69 @@
...
+*            }
+*         ]
+* @return {void} Nothing
+*/
+async function dispatch(lib, data) {
+  const { mu, muAuthSudo } = lib;
+  const { termObjects } = data;
+
+  // Steps:
+  // 1. Write the triples to the landing zone graph - withouth any mapping or filtering
```
mirdono, line 33: **Typo:** "withouth" should be "without"

docker-compose.yml
```diff
@@ -624,6 +624,37 @@ services:
       - triplestore:virtuoso
     logging: *default-logging
 
+################################################################################
+  # CLB CONSUMERS
+  ################################################################################
+  clb-contact-data-consumer:
+    image: lblod/delta-consumer:0.0.17-rc.7
```
mirdono, line 631: **Question:** I noticed we are currently using a mix of 0.0.17-rc.7 and 0.0.18 of the `delta-consumer` service. Is there a reason you opted for this version?

mirdono (commented):
Some comments based on a pass over the code. I did not run it.

docker-compose.yml
lagartoverde: Not really I just copied another consumer config, I updated it to 0.0.18 as it seems to be a newer version without any breaking changes

mirdono (commented):
Some additional remarks:
- the website of an organisation is not received, I think because CLB's [producer configuration](https://github.com/lblod/app-contactgegevens-loket/blob/ac6cdef6eeb2aab698b7d3557e21eca8e2c33fd7/config/delta-producer/contact-data/export.json#L17C17-L17C42) contains a different predicate than the [resources configuration](https://github.com/lblod/app-contactgegevens-loket/blob/ac6cdef6eeb2aab698b7d3557e21eca8e2c33fd7/config/resources/domain.json#L286)
- I am not sure whether I am missing a piece of the puzzle here, but on my setup the data arriving from CLB stops at the landing graph, and is not further propagated.

lagartoverde:
So I rechecked everything, maybe the data not arriving to the correct graph was because we had an incorrect dump on CLB due to timing issues.
But also you have to be careful because of the difference between environments there's some admin units that are in CLB dev and not in your local machine, this data won't go to the administrative unit graph because it doesn't find an admin unit to link to. I tested with AGB Tienen who is both in CLBV dev and the local migrations, there are more, but check the uris are the same

mirdono (commented):
I tested with AGB Tienen, as you suggested, and editing the existing site works mostly as expected.
Three observations though:
- When I added a new site for AGB Tienen in CLB, the deltas do arrive at OP but are not properly processed. The triples are not added to the landing zone graph (or are added but removed again, based on my analysis of the consumer logs I am not sure), and the site does not show up in the administrative unit graph. Doing a new initial sync also fails to pick up the new site.
- Changing the value for the primary correspondence address ("Primair correspondentieadres") in CLB does not seem to be synced to OP. For example, again for AGB Tienen, I changed the primary correspondence address to the newly added site but this change does result in any delta's arriving at OP.
- The optional name for sites of type "Andere vestiging" is not send to OP. Changing this in CLB does not result in deltas arriving at OP, and its value is not included in delta's received after other changes. (Sidenote, the label for that name field should probably be the Dutch "Naam" instead of the English "Name".)

lagartoverde:
> I tested with AGB Tienen, as you suggested, and editing the existing site works mostly as expected. Three observations though:
> 
> * When I added a new site for AGB Tienen in CLB, the deltas do arrive at OP but are not properly processed. The triples are not added to the landing zone graph (or are added but removed again, based on my analysis of the consumer logs I am not sure), and the site does not show up in the administrative unit graph. Doing a new initial sync also fails to pick up the new site.
> * Changing the value for the primary correspondence address ("Primair correspondentieadres") in CLB does not seem to be synced to OP. For example, again for AGB Tienen, I changed the primary correspondence address to the newly added site but this change does result in any delta's arriving at OP.
> * The optional name for sites of type "Andere vestiging" is not send to OP. Changing this in CLB does not result in deltas arriving at OP, and its value is not included in delta's received after other changes. (Sidenote, the label for that name field should probably be the Dutch "Naam" instead of the English "Name".)

Hello Alexander thanks for your review, the new sites (and the change of normal to primary site because they work the same way) was not working because of a bug in the consumer, I fixed that and tested that it correctly adds new sites and proccess the changes, the name of the site type was not added because it was added after the producer was created, I added it to CLB, but it will take some time to reach dev, and you are right about the name I will propose a change on CLB

mirdono (commented):
Some observations based after testing (as before with AGB Tienen)
- For new sites added on CLB the `uuid` predicate is not added to the store on OP's side. So none of the new entities (site, address, and contact points) have a UUID and are therefore not shown in OP's frontend. A look at the consumer's docker logs seems to indicate the `uuid` predicate is not received from CLB at all. The other triples are received and processed mostly correctly. (To be able to test further I manually inserted the necessary UUIDs in my local store.)
- In some cases the changed triples are added next to the already existing ones instead of overwriting them. I noticed this, for example, when editing address information on CLB after which the changed information, in this case street and number, was added to the correct graph in OP but the old values were not removed. Unfortunately I am unable to reproduce this consistently. For instance, changing the same values again does work as expected in that the previously set values are deleted.  Maybe only the first edit after creating a new site goes wrong?
- When manually specifying an address without a bus number (busnummer) in CLB an empty string is inserted as value on OP's side. AFAIK we tend to not write such triples to our store. 
- Not sure if this is expected behaviour, but I noticed that the queries to insert and delete data specify the same data multiple times. For example, the following queries were performed by the consumer when changing a single phone number:
```
DELETE DATA {
    GRAPH <http://mu.semte.ch/graphs/administrative-unit> {
      <http://data.lblod.info/id/contact-punten/660D71BD5C8A6B07BE99ED68> <http://schema.org/telephone> """tel:+3211211000""".
      <http://data.lblod.info/id/contact-punten/660D71BD5C8A6B07BE99ED68> <http://schema.org/telephone> """tel:+3211211000""".
    }
   }

    INSERT DATA {
      GRAPH <http://mu.semte.ch/graphs/administrative-unit> {
        <http://data.lblod.info/id/contact-punten/660D71BD5C8A6B07BE99ED68> <http://schema.org/telephone> """tel:+3211211555""".
        <http://data.lblod.info/id/contact-punten/660D71BD5C8A6B07BE99ED68> <http://schema.org/telephone> """tel:+3211211555""".
      }
    }
```

docker-compose.yml
```diff
@@ -624,6 +624,37 @@ services:
       - triplestore:virtuoso
     logging: *default-logging
 
+################################################################################
+  # CLB CONSUMERS
+  ################################################################################
+  clb-contact-data-consumer:
+    image: lblod/delta-consumer:0.0.18
```
mirdono, line 631: **Suggestion:** Bump to v0.0.24 as we did for the other consumers recently in #401.

mirdono (approved):
My previous comments were fixed.

PR #384, opened 2024-02-27
Chore: Add migration template samples

aliokan:
Copy the `migrations-sample` and `migrations/local` logic in place in Lokaal Beslist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ docker-compose.override.yml
 .env
 
 /config/migrations/private-data/*
+/config/migrations/local/*
 
 /config/proxy/nginx.htpasswd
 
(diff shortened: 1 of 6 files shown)


mirdono (commented):
Could you also take a look at the extra, unversionned migrations on PROD? I think at least the migration to handle the occasional sync issue between OP and Loket (`20231214114200-remove-type-in-producer-graph.sparql` on PROD) would be a useful template to have.

Otherwise it looks fine to me.

aliokan:
@mirdono template added

mirdono (approved):

mirdono:
Apologies for adding stuff after approving but on QA there is also a migration `00000000000000-rm-stale-jobs.sparql` that seems relevant to add here.

PR #385, opened 2024-02-28
Feature: add legal name for organisations

mirdono:
OP-2999

# Summary
- Add a `legal-name` attribute to organisations
- For existing organisations, initialise this new attribute with the value of `skos:prefLabel` via a migration
- Include this new attribute in the search index

# TODO
- [x] Prevent data from being duplicated to `administrative-unit` graph when altering legal name for a municipality or province


aliokan (commented):
@mirdono I left one question

aliokan (approved):
:rocket:

PR #387, opened 2024-03-04
Exclude the data of CLB from the producer

lagartoverde:
OP-2982


mirdono (commented):
Looks good to me, not an expert on producers though.

aliokan (commented):
**Question:** Is only CLB use that producer?

lagartoverde:
> **Question:** Is only CLB use that producer?
Yes

lagartoverde:
Can I merge this then?

aliokan (approved):
:+1:

PR #389, opened 2024-03-06
fix: set correct value for provinces of OCMW associations and PEVAs

mirdono:
OP-3062 and OP-3063

## Context
In the initial migrations the province was incorrectly set to the same value as the municipality for OCMW associations as wll as PEVAs. These new migrations first remove the incorrect province and then insert the correct value.

## Notes
- These migrations are generated using a script so each query is identical except for the concrete values such as KBO number and province. It is thus not necessary to go through all the queries in the migrations.
- The province column on the index page will still display the old, incorrect values. Updating these requires a reindexing. The province shown on the core data page should display the new, corrected value.


Dachamelemon (approved):
Migrations OK!

PR #390, opened 2024-03-07
Feature: add alternative names for organisations

mirdono:
OP-3000

## Summary
- Changed type for `alternative-name` attribute to `string-set` which allows to retrieve multiple values for the same predicate
- Add alternative name attribute to search index
- Fronted [PR](https://github.com/lblod/frontend-organization-portal/pull/586)


aliokan (approved):
Looks good! :rocket:

PR #391, opened 2024-03-11
Add administrative-unit in delta-producer/organisations

aliokan:
Related to OP-2957, OP-3128

This PR contains modifications required by Kalliope consumer.  
- add `http://data.vlaanderen.be/ns/besluit#Bestuurseenheid`
- add website (`http://xmlns.com/foaf/0.1/page`) in ContactPoint
- add `http://www.w3.org/ns/prov#Location` (the glue between administrative-unit with NISCODE)

### DIFF
--- a/config/delta-producer/organizations/export.json
+++ b/config/delta-producer/organizations/export.json
@@ -353,6 +353,18 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


mirdono (approved):
Looks good to me.

PR #396, opened 2024-03-15
Feature: add merge change event pz aalter maldegem

mirdono:
OP-3067

## Notes
- The resulting administrative unit "PZ Aalter/Maldegem" is only found in PROD data.


aliokan (approved):
Looks good, thanks for the comments! :pray:

PR #401, opened 2024-03-29
chore: bumped consumers to latest version

mirdono:
Bumped the 3 consumers to the latest version of the `delta-consumer`. It should be backwards compatible but some testing/monitoring seems prudent.


aliokan (approved):
:rocket:

PR #406, opened 2024-04-04
Add legal and alternative names to producer

lagartoverde:
In order to show them in CLB


mirdono (changes requested):
For most types the `skos:altLabel` was already included, and is now duplicated. Other than that it looks fine to me

mirdono (approved):
:+1:

PR #408, opened 2024-04-04
feat: export legal and alternative names via producers

mirdono:
Following #406 this adds the legal and alternative names to the exports for all other consumers.

Notes:
- The alternative name was previously already included in most exports.
- I do not think the consumers of our data require immediate changes to keep up to date data. The frontend will update the value for `skos:prefLabel` whenever a user changes an organisation's `regorg:legalName`.


aliokan (approved):
Good catch! :pray:

PR #409, opened 2024-04-09
removed empty literals and uris

Dachamelemon:
All removed empty literals and uris.

FOr the URIs there are still a few empty ones in the landingzone and producer graphs. I'm not entirely sure if i need to include them.
Comparing to a similar query done in the past it seems that they included those too but I am not sure if that was just an irregularity from back then or not

### DIFF
--- a/config/migrations/2024/20240316100500-delete-all-empty-uris.sparql
+++ b/config/migrations/2024/20240316100500-delete-all-empty-uris.sparql
@@ -15,7 +15,8 @@ WHERE {
   GRAPH ?g {
     ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?type  ;
       <http://mu.semte.ch/vocabularies/core/uuid> ?uuid .
-
+      
+    FILTER (?type != <https://data.vlaanderen.be/ns/generiek#GestructureerdeIdentificator>)
     FILTER NOT EXISTS {
       FILTER (?p != <http://www.w3.org/1999/02/22-rdf-syntax-ns#type>)
       FILTER (?p != <http://mu.semte.ch/vocabularies/core/uuid>)
--- a/config/migrations/2024/20240316100500-delete-all-empty-uris.sparql
+++ b/config/migrations/2024/20240316100500-delete-all-empty-uris.sparql
@@ -1,0 +1,25 @@
+DELETE {
+  GRAPH ?g {
+    ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?type  ;
+      <http://mu.semte.ch/vocabularies/core/uuid> ?uuid . 
+  }
+}
+WHERE {
+  VALUES ?g {
+    <http://mu.semte.ch/graphs/administrative-unit>
+    <http://mu.semte.ch/graphs/worship-service>
+    <http://mu.semte.ch/graphs/public>
+    <http://mu.semte.ch/graphs/privacy-centric-graph>
+    <http://mu.semte.ch/graphs/worship-privacy-centric-graph>
+  }
+  GRAPH ?g {
+    ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?type  ;
+      <http://mu.semte.ch/vocabularies/core/uuid> ?uuid .
+
+    FILTER NOT EXISTS {
+      FILTER (?p != <http://www.w3.org/1999/02/22-rdf-syntax-ns#type>)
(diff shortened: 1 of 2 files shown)


config/migrations/2024/20240316100500-delete-all-empty-uris.sparql
```diff
@@ -0,0 +1,25 @@
...
+    ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?type  ;
+      <http://mu.semte.ch/vocabularies/core/uuid> ?uuid .
+
+    FILTER NOT EXISTS {
+      FILTER (?p != <http://www.w3.org/1999/02/22-rdf-syntax-ns#type>)
+      FILTER (?p != <http://mu.semte.ch/vocabularies/core/uuid>)
+      ?s ?p ?o .
+    }
+  }
+}
```
mirdono, line 25: **Question:** Have you considered the situation where `?s` also occurs as an object in some other triple? In other words what happens when there exists a triple `?s1 ?p1 ?s`, for which `?s`  satisfies the body in the `WHERE` clause and would thus be deleted.

A more concrete example. When you create a new organization **without** a SharePoint identifier. The necessary placeholder triples for a future identifier are already created. The result would look something like this:

| ?s | ?p | ?o |
| -- | -- | -- |
| <http://data.lblod.info/id/bestuurseenheden/661689F0315E0C916A052AB2> | <http://www.w3.org/ns/adms#identifier> | http://data.lblod.info/id/identificatoren/661689EE315E0C916A052AAD> |
| <http://data.lblod.info/id/identificatoren/661689EE315E0C916A052AAD> | <http://www.w3.org/2004/02/skos/core#notation> | "SharePoint identificator" |
| <http://data.lblod.info/id/identificatoren/661689EE315E0C916A052AAD> | <http://mu.semte.ch/vocabularies/core/uuid> | "661689EE315E0C916A052AAD" |
| <http://data.lblod.info/id/identificatoren/661689EE315E0C916A052AAD> | <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> | <http://www.w3.org/ns/adms#Identifier> |
| <http://data.lblod.info/id/identificatoren/661689EE315E0C916A052AAD> | <https://data.vlaanderen.be/ns/generiek#gestructureerdeIdentificator> | <http://data.lblod.info/id/gestructureerdeIdentificatoren/661689EE315E0C916A052AAC> |
| <http://data.lblod.info/id/gestructureerdeIdentificatoren/661689EE315E0C916A052AAC> | <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> | <https://data.vlaanderen.be/ns/generiek#GestructureerdeIdentificator> |
| <http://data.lblod.info/id/gestructureerdeIdentificatoren/661689EE315E0C916A052AAC> | <http://mu.semte.ch/vocabularies/core/uuid> | "661689EE315E0C916A052AAC"

The `gestructureerdeIdentificatoren` triples on the bottom two rows seem to satisfy the `WHERE` clause in the migration's query. Yet we do not want these to be deleted as we need them when an SharePoint identifier is added at some point. (That action will add extra triples relating the structured identifier resource to a new local identifier resource, which in turn has a triple with the actual identifier value as literal.)

config/migrations/2024/20240327131000-delete-all-empty-literals.sparql
```diff
@@ -0,0 +1,20 @@
...
+    <http://mu.semte.ch/graphs/administrative-unit>
+    <http://mu.semte.ch/graphs/worship-service>
+    <http://mu.semte.ch/graphs/public>
+    <http://mu.semte.ch/graphs/privacy-centric-graph>
+    <http://mu.semte.ch/graphs/worship-privacy-centric-graph>
+  }
+  graph ?g {
+    ?s ?p ?o
+
+    Filter(?o = " " || ?o = "")
```
mirdono, line 17: **Question:** Are we sure that a single space and the empty string are the only possibilities? Would it be worth it to filter based on a regex by replacing the filter by (warning untested!!) `FILTER regex(str(?o), "^\\s*$")`.
Note, adding an additional `FILTER (isLiteral(?o))` before the regex filter might be necessary for performance.

mirdono (commented):
@Dachamelemon I left some questions

Dachamelemon:
closed because removing the empty uris will be discussed later to see what it could break. Only the removal of empty literals will be done see PR: https://github.com/lblod/app-organization-portal/pull/413

PR #414, opened 2024-04-17
Fix consumer problem with admin units in shared graph

lagartoverde:


aliokan (commented):
@lagartoverde Could you add a commentary to explain why you add the `shared` graph, for you or the next one that have to edit this config file in 6 months ;)

aliokan (approved):
@lagartoverde Thank you for the comments! :pray:

PR #416, opened 2024-04-23
avoid inserting duplicate triples to the database

lagartoverde:
This lowers down the problematic delta from 200k insertions to "only" 2k. Still takes a lot of time because of the cascade queries we make, but there's no solution for that yet according to Boris.
Also this deltas will be removed in the future from CLB as they are an error so we won't have this problem, but anyway this makes the consumer more perfomant overall


aliokan (approved):
I haven't tested, but the code looks good! :+1:

PR #418, opened 2024-04-23
Refactor: generalise backend for non-administrative unit organisations

mirdono:
OP-3142

Corresponding frontend: [PR 602](https://github.com/lblod/frontend-organization-portal/pull/602)

## Changes

- Moved the classification relationship from `administrative-units` to `organizations` and generalised its target to `organization-classification-codes`.
- Rename the `units` index in the search configuration to `organizations` to indicate it covers more than administrative units. (Which it already did since it includes representative bodies.)
- Dispatch requests on the `organizations` path directly to the `resources` service instead of the `cache` service

### DIFF
--- a/config/dispatcher/dispatcher.ex
+++ b/config/dispatcher/dispatcher.ex
@@ -73,7 +73,7 @@ defmodule Dispatcher do
   end
 
   match "/organizations/*path", %{ accept: [:json], layer: :api} do
-    Proxy.forward conn, path, "http://cache/organizations/"
+    Proxy.forward conn, path, "http://resource/organizations/"
   end
 
   match "/administrative-units/*path", %{ accept: [:json], layer: :api} do
--- a/config/resources/domain.json
+++ b/config/resources/domain.json
@@ -50,9 +50,7 @@
     },
     "people": {
       "class": "person:Person",
-      "super": [
-        "agents"
-      ],
+      "super": ["agents"],
       "attributes": {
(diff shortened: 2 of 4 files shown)


config/dispatcher/dispatcher.ex
```diff
@@ -73,7 +73,7 @@ defmodule Dispatcher do
   end
 
   match "/organizations/*path", %{ accept: [:json], layer: :api} do
-    Proxy.forward conn, path, "http://cache/organizations/"
+    Proxy.forward conn, path, "http://resource/organizations/"
```
aliokan, line 76: As discuss: 
- [ ] revert and check if it's working after restart `resource` and `cache` services

mirdono: With restarting the cache it does seem to work as expected without this change. I removed the commit introducing this change.

mirdono: I ~~spoke~~ wrote to soon apparently. I encountered an issue with using the cache path. When you edit, for instance, the related organizations for an organization the results shown after saving are not updated and keep showing the state from before editing. Changing the path to resource to bypass the cached results resolves this issue.

aliokan (approved):
@mirdono I will not have the time before the end of the sprint to understand why the cache is not working correctly. 
This cache issue shouldn't block the PR, I have open a follow-up ticket (OP-3173) to manage it

PR #419, opened 2024-04-24
fix/change-event-date-update-migration

Dachamelemon:
fixed issue with dates in migration


mirdono:
**Question:** The original ticket only concerns worship services. Is there a reason to include the `administrative-unit` graph as possible target?

Dachamelemon:
> **Question:** The original ticket only concerns worship services. Is there a reason to include the `administrative-unit` graph as possible target?

Not really I will take it out.

mirdono (commented):
**Issue:** There are some problems with the change events concerning the worship service with UUID `a8d9f1a45c11c6761306bdf680349593` (line 96 in the excel). Going to the change events page for this organization results in an error in the console, which is already the case before this migration. The result of the migration here does seem to duplicate the change event in question, it is shown twice in the frontend. Could you look into this?

**Nitpick:** Could you update the bug ticket (OP-3160)?

mirdono (approved):
As discussed during today's stand-up (25/04), let's move forward with replacing the previous migration with this improved version for better data consistency. The mentioned bug concerning a specific worship service will be handled separately.

PR #420, opened 2024-04-29
fix: remove dangling change event results

mirdono:
OP-3168

The change events related to an organisation contains some data errors:
- There is a change event result without any corresponding change event. This
  causes an error when visiting the change events page for this organisation.
- There are two identical change event results for the same change event. This
  causes the change event to be shown twice in the table. (Only on PROD)

The added migration removes these unnecessary elements change event results.

Note: it may be required to also restart the cache service before you correctly
see the result of this migration. Without a restart the same error and duplicate
entries kept popping up in the frontend for me.


aliokan (approved):
@mirdono Yes, restart the cache is mandatory, as mu-search, the cache need to be notified when a change occurred and there's no notification involved with migrations.

PR #423, opened 2024-05-17
feat: import participants for public OCMW associations

mirdono:
OP-3040

## Notes
- The used business data can be found in the excel linked in OP-3037. This PR only concerns the data in the "VVMV_participants_public" sheet.
- The migration was generated using a python script that can be found in the ZIP file in a comment in ticket OP-2540. The script is located in the subfolder "05-related-organisations".
- The migration for "Ter Lembeek" was manually adapted from the generated one as the business data does not contain a KBO number for this organisation.


aliokan (approved):
I didn't check the business requirement, but queries looks good :+1:

PR #425, opened 2024-05-22
fix: exclude KBO organisations as participants

mirdono:
The initial version of these migrations introduced in #423 were defined too
broadly. Since they only matched on KBO numbers they also included the KBO
organisations, imported from wegwijs, adding them as participants.

This PR fixes these issues by
- Removing the previously imported data using a DEV-specific migration
- Add additional `FILTER` statements to exclude KBO organisations
- Split the main migration into two parts to work around an error otherwise
  encountered when executing the migration in one go


aliokan (approved):
:+1:

PR #431, opened 2024-06-05
Delete duplicate identifier on organization

aliokan:
Related to OP-3157

## Context

For some reason, we have duplicate values attached to the same organization. 

## Proposal

Delete all duplicates with a migration


mirdono (approved):
Looks good to me. One word of warning, the migration crashed once for me locally. So you might want to keep an eye on the logs when deploying.

PR #433, opened 2024-06-07
Disable cache on `/kbo-organizations/*`

aliokan:
We cannot use cache on `kbo-organizations` route, because data are updated by `lblod/sync-wegwijs-organization-service` and the cache will be never invalidated.


mirdono (approved):
Good catch!

PR #435, opened 2024-06-19
fix: remove `org:memberOf` triples with KBO organizations as object

mirdono:
OP-3286

The migration importing related organizations for public OCMW associations was
incorrect in that it did not exclude resources of type `ext:KboOrganisatie`
being matched. This migration removes all incorrectly inserted triples.

The incorrect migration was originally added in #423 and an initial fix for the
same issue was proposed in #425. The fix was also incorrect as it added a filter
on the participant resource instead of the association resource, which is the
resource that can be of the incorrect type.


aliokan (approved):
:+1:

PR #436, opened 2024-06-20
fix: import participating municipalities for public OCMW associations

mirdono:
OP-3288

The original migration to import the participants for public OCMW associations
(see #423 and #425) contained a bug causing it to not match municipalities. More
specifically:
- the type triple for municipalities is located in the `shared` graph;
- since the migration's queries are limited to the `administrative-unit` graph
  the `FILTER EXISTS` statements to limit matches to administrative units could
  not find a match for municipalities.

This new migration fixes this problem by changing the `FILTER EXISTS` into a
`FILTER NOT EXISTS` to exclude resources of type `ext:KboOrganisatie`. For
municipalities the `ext:KboOrganisatie` resource is fully stored in the
`administrative-unit` graph.


aliokan (approved):
:crossed_fingers:

PR #440, opened 2024-07-18
fix: combine two merger change event

mirdono:
OP-3298


Riadabd (approved):
Tested locally, and the migration adds and removes the expected data :+1:

PR #447, opened 2024-08-28
export ro classifcation in other public producer too

bdevloed:
add classification to public producer used by loket (previous was org-public)


mirdono (commented):
While we are at it, should the `org:classification` property also be added to the [organizations](https://github.com/lblod/app-organization-portal/blob/90cc3a088879eaefa598650227c2401c6d664032/config/delta-producer/organizations/export.json#L212) and [worship-posts](https://github.com/lblod/app-organization-portal/blob/90cc3a088879eaefa598650227c2401c6d664032/config/delta-producer/worship-posts/export.json#L119) producers? They currently both produce representative bodies without classification.

PR #451, opened 2024-09-26
Add queries to remove organizations

Riadabd:
## Ticket ID

OP-3124 + OP-3125

## Ticket Description

This PR deletes organization URIs (+ other related URIs) as this data is deemed to be unneeded.

## How to test

The best way to test this would be:
* Tunnel to production, use `DESCRIBE <bestuurseenheid_URI>` and follow the links.
* Do a quick skim between the URIs connected to the bestuurseenheid URI and cross-check with the queries.

The cross-checking part is a bit annoying, but there aren't many resources attached to these organizations, so it should be fairly quick.

### DIFF
--- a/config/migrations/2024/20240916161553-remove-organizations/20240916161554-remove-limgrond-be-data.sparql
+++ b/config/migrations/2024/20240916161553-remove-organizations/20240916161554-remove-limgrond-be-data.sparql
@@ -1,0 +1,129 @@
+DELETE {
+  GRAPH ?g {
+    ?s ?p ?o .
+  }
+}
+WHERE {
+  VALUES ?s {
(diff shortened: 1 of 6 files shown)


claire-lovisa (changes requested):
I think that the bestuursfuncties (accessible via `?orgaanInTime <http://data.lblod.info/vocabularies/leidinggevenden/heeftBestuursfunctie> ?bestuursfunctie` should also go because they are specific to those organizations

Riadabd:
> I think that the bestuursfuncties (accessible via `?orgaanInTime <http://data.lblod.info/vocabularies/leidinggevenden/heeftBestuursfunctie> ?bestuursfunctie` should also go because they are specific to those organizations

Thanks for the comment! I was wondering if those should be deleted or not; I'll update my migrations.

claire-lovisa (approved):

PR #454, opened 2024-10-07
fix: make memberships compatible with produced data

mirdono:
## Summary
Other apps do not (yet) use memberships to model relations between organisations. Therefore, OP must ensure its produced data remains compatible despite internally using memberships. The proposed solution is to use the [db-cleanup-service](https://github.com/lblod/db-cleanup-service) to periodically insert/delete the necessary triples.

## Proposed
In order to remain compatible OP needs to produce triples for 2 predicates, `org:hasSubOrganization` and `org:linkedTo`[^org], for "has relation with" memberships.

### Migrations introducing memberships
The migrations that introduce memberships have been update to **not** delete triples for the above predicates.
Note, on DEV these migrations have already been executed and the triples have already been removed. This will be automatically resolved by the cleanup jobs described below.

### DB cleanup service
The DB cleanup service is configured to periodically run 10 jobs, 5 for inserting additional triples for existing memberships and 5 for deleting triples that do no longer match a membership.

For the following "has a relation with" memberships an additional triple should be inserted if it does not exist yet:

| Organization classification | Member classification | Corresponding triple                         |
|-----------------------------|-----------------------|----------------------------------------------|
| Worship service             | Representative body   | ?member org:linkedTo ?organization           |
| Central worship servive     | Representative body   | ?member org:linkedTo ?organization           |
|                             | Worship service       | ?organization org:hasSubOrganization ?member |
| Municipality                | APB                   | ?organization org:linkedTo ?member           |
|                             | OCMW                  | ?organization org:linkedTo ?member           |
|                             | AGB                   | ?organization org:hasSubOrganization ?member |
|                             | Assistance zone       | ?organization org:hasSubOrganization ?member |
|                             | IGS                   | ?organization org:hasSubOrganization ?member |
|                             | PEVA municipality     | ?organization org:hasSubOrganization ?member |
|                             | Police zone           | ?organization org:hasSubOrganization ?member |
| Province                    | AGB                   | ?organization org:hasSubOrganization ?member |
|                             | APB                   | ?organization org:hasSubOrganization ?member |
|                             | OCMW                  | ?organization org:hasSubOrganization ?member |
|                             | PEVA                  | ?organization org:hasSubOrganization ?member |

In the last column`?organization` and `?member` refer to the links between the membership and its involved organisations. Thus the listed triples would correspond to the following membership:

``` sparql
?membership a org: Membership ;
  org:organization ?organization ;
  org:member ?member ;
[...]

PR #452, opened 2024-10-03
Bump delta producers and move to publication triplestore

Riadabd:
## Ticket ID

OP-3422 (*Related ticket:* OP-2883 => This **should** happen during the deploy of this PR; a [couple of docs](https://github.com/Riadabd/dump-and-import-publication-graphs/tree/main/docs/app-organization-portal) have been made to assist with this process)
  - Should fix DL-6192

## Description

This PR moves individual delta streams (`public`, `public-info` and `worship-posts`) to the consolidated `delta-producer-publication-graph-maintainer` in addition to moving these stream's background jobs initiator configs to the general `delta-producer-background-job-initiator`.

## How to Test

The main point to test (and the motivator for this PR) is the representative bodies issue that has been surfacing for a while, but there may be other stuff that need to be tested to make sure each individual stream is still working as expected.

### RO Issue

1. Run the app without this branch (use the dev server database).
2. Go to `Josephine BothRoles` -> `Erediensten - Editeerder` -> `Ga naar organisaties` -> `+ Nieuw`.
3. Fill the data as in the pic below:
image
4. You will be taken to the following URL: `http://localhost/organisaties/UUID/kerngegevens`.
5. Use the UUID in the link to fetch the organization's URI.
6. Run `DESCRIBE <ORG_URI>` and fetch the RO pointing to this organization.
7. Once you fetch the RO URI, run the following query in `publication-triplestore`:
```sparql
SELECT ?p ?o WHERE {
  GRAPH <http://redpencil.data.gift/id/deltas/producer/public> {
    <RO_URI> ?p ?o .
  }
}
```
8. You should see some data there (e.g., `prefLabel`, `identifier`, `uuid`, `type`, etc..)
9. Go back again to `http://localhost/organisaties/UUID/kerngegevens` and click on `Bewerk`.
10. Edit the `Juridische naam` field.
11. Inspect the logs of `delta-producer-pub-graph-maintainer-public` and wait a couple of minutes.
12. Refresh the browser tab where you have the SELECT query -> Most data related to the RO in the `http://redpencil.data.gift/id/deltas/producer/public` graph should be gone.

Checkout the PR now and perform the following:
1. `rm -rf data/db && rm -rf data/publication-triplestore && sudo rm -rf data/elasticsearch`
2. Copy the dev db again to `data/`.
3. Up the stack and run the healing for the `public` stream in order to populate `publication-triplestore` (since it doesn't contain useful data yet) (this part may take some time since the service has to generate the delta files as well):
```
[...]

PR #456, opened 2024-10-11
Link missing classifications to public producer concept scheme

claire-lovisa:
# Context

OP-3435
The bestuursfuncties were not produced (as planned in the export configuration) because of a missing link to the export concept scheme of the public producer. This migration restores it.

# How to test it / deploy it

First we need to run the migration. Once this is done, by triggering (or waiting for) a healing of the public producer, we should see the bestuursfuncties arriving in the producer's graph.


claire-lovisa:
@mirdono is there a realease planned soon-ish where this could be embedded ? Is it OK to merge it to dev or better merge into master ?

mirdono:
> @mirdono is there a realease planned soon-ish where this could be embedded ? Is it OK to merge it to dev or better merge into master ?

The next release is v1.28 which already contains quite a lot: memberships, ember update, bunch of smaller stuff. This release is currently blocked from going into pre-release until the functionality added in #454 has been approved by the testing team (which hopefully can be done in the current sprint). In order to avoid piling more on that release as well as avoiding delays in deploying this PR (I expect v1.28 to be in pre-release stage for a while), I would be inclined to merge to master and release this as a hotfix together with #452 .

PR #474, opened 2024-10-30
Hotfix/export vgc to public

cecemel:
VGC data was missing to public export. Adding it here.


mirdono (approved):
Looks correct. Just a sidenote, OP does not show the VGC's governing bodies to the user (this was not a requirement at the time of adding this organisation), so unless this is changed the imported information will not be visible/editable to the users.

PR #475, opened 2024-10-30
Added config export to export ABB

cecemel:
ABB should also be exported as bestuurseenheid.


mirdono (commented):
Looks good to me, barring if you still want to add a classification for that organisation.

PR #478, opened 2024-11-26
chore: rename municipalities and OCMWs for mergers 2025

mirdono:
As part of the 2025 municipality mergers 3 municipalities, and their
corresponding OCMWs, keep their original URI but do change their name. More
specifically:

| Current name | Name after merger |
| ------------ | ----------------- |
| Bilzen       | Bilzen-Hoeselt    |
| Ham          | Tessenderlo-Ham   |
| Tongeren     | Tongeren-Borgloon |

## Related tickets
- OP-3456


claire-lovisa (approved):
Very handy those migrations triggering indexing :grin:

mirdono:
> Very handy those migrations triggering indexing 😁

Idd, just a bit unfortunate it does not allow comments in the migrations due to being parsed by the authz service.

claire-lovisa:
should be better once we move to sparql-parser instead of mu-auth from what i remember :crossed_fingers:

mirdono:
Doing that move has been on my TODO list since the sparql-sparser presentation a bit ago. Hopefully I get around that soon :crossed_fingers:

PR #479, opened 2024-11-26
chore: bump services

mirdono:
Bump the following services to the most recent versions:
- identifier to v1.10.3
- migrations and migrations-triggering-indexing to v0.9.0
- resource to v1.25.0
- file to v3.4.0

The versions bumped to do not introduce breaking changes or changes requiring additional changes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,21 +51,21 @@ services:
       - "logging=true"
     restart: always
     logging: *default-logging
+  db-cleanup:
+    image: lblod/db-cleanup-service:0.5.1
+    environment:
+      MU_SPARQL_ENDPOINT: "http://db:8890/sparql"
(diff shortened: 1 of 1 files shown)


claire-lovisa (changes requested):
I can't access the core data of admin units, mu-cl-resource is throwing an error:
```
resource-1                                           | An unhandled error condition has been signalled:
resource-1                                           |    There is no applicable method for the generic function
resource-1                                           |      #<STANDARD-GENERIC-FUNCTION MU-CL-RESOURCES::RESOURCE-NAME (3)>
resource-1                                           |    when called with arguments
resource-1                                           |      (NIL).
```
Do you have it as well?

claire-lovisa (approved):
I was on the wrong frontend branch, it now works well!

PR #481, opened 2024-12-03
feat: set end date governing bodies for merged organisations

mirdono:
Set the end date for governing bodies for municipalities and OCMWs that are
involved in a municipality merger **and** will become inactive afterwards.

For the municipalities and OCMWs that remain active after the merger, the end
date has already been set previously.

## How to test
- Run the added migration
- Via the frontend:
    + Go to the governing bodies page for a relevant organisation
      (they are listed in the `VALUES` clause in the migration's query)
    + Check whether the end date for the governing bodies that start in 2019 is
      set to 02/01/2025

## Notes
- The status of the governing bodies is automatically determined in the frontend
  based on the start and end dates. So no explicit status changes are required.
- OP-3412 contains an overview of which organisations are involved in mergers
  and which ones remain/become active or become inactive

## Related tickets
- OP-3348


claire-lovisa (approved):
Looking good!

PR #485, opened 2024-12-11
Fix skos:notation ADMS:identifier language tags

cecemel:
It seems there were language tagged and non-languaged versions used for the skos:notation from ADMS:identifier.
This migrations alings it, we don't want typed strings here. See also DL-6343


mirdono (approved):
Looks good to me

PR #486, opened 2024-12-12
Archive AGB Stekene

Riadabd:
## Ticket ID

OP-3409

## Description

AGB Stekene has be to archived; we set its org status to `Niet-Actief`.


Riadabd:
Is setting `orgStatus` to `Niet-Actief` enough to archive the organization in OP? Or are there extra steps that need to be taken there?

mirdono:
> Is setting `orgStatus` to `Niet-Actief` enough to archive the organization in OP? Or are there extra steps that need to be taken there?

To my knowledge this is the only thing you can do, aside from deleting the organisation, to "archive" an organisation.

mirdono (approved):
**Suggestion:** You can use the `migrations-triggering-indexing` service instead of the regular `migrations` service. That way you do not have to run a full re-index just to have one status updated. You will have to remove the `# "Niet-Actief" Status` comment in the query, as the `mu-authorization` service cannot handle comments.

cecemel:
@mirdono @Riadabd I updated the migration accordingly. So it's going to be a ` drc restart migrations-triggering-indexing`.  I don't know how to not forget this on deploy.

PR #487, opened 2024-12-12
Remove empty URIs

Riadabd:
## Ticket ID

OP-3426

## Description

The Kalliope team notified us of empty URIs they received; these URIs only have `http://www.w3.org/1999/02/22-rdf-syntax-ns#type` and `http://mu.semte.ch/vocabularies/core/uuid` properties attached to them. So, they can be safely deleted.

## How to test

Run the migration and make sure these URIs no longer have any data.


mirdono (approved):
Looks good.

**chore:** In the process of reviewing I also came across `<http://data.lblod.info/id/bestuurseenheden/3183ade3-7ee8-409c-8394-296b1fbfe478>` for which only a single type triple exists (in PROD). While not explicitly listed in the ticket this can probably be removed along with the others in this migration.

cecemel:
Updated the migration.

PR #493, opened 2025-01-08
Migration to delete betrokkene and update financieringspercentage

wolfderechter:
## ID
OP-3501

## Description

Remove the betrokken bestuur of St-Amandus, 'Oudenaarde'. And update the financierings percentage to 100 for the other betrokkene 'Wortegem-Petegem'.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

## How to test

Rsync prod data, run migration and verify that the betrokken bestuur (Oudenaarde) is removed, and the other betrokkene (Wortegem-Petegem) has a percentage of 100.

### DIFF
--- a/config/migrations/2025/20250107101243-delete-local-involvement-oudenaarde-st-amandus.sparql
+++ b/config/migrations/2025/20250107101243-delete-local-involvement-oudenaarde-st-amandus.sparql
@@ -1,0 +1,35 @@
+DELETE WHERE {
+  graph <http://mu.semte.ch/graphs/worship-service> {
+    <http://data.lblod.info/id/betrokkenLokaleBesturen/0da11d9fdef2d45f90fd4ebd353ecd3c> ?p ?o.
+  }
+}
+
+;
+
+DELETE WHERE {
+  graph <http://mu.semte.ch/graphs/worship-service> {
+    <http://data.lblod.info/id/bestuurseenheden/d9f7c0ab4920fdecf3f9a60b92e921b5ca07248fcb0eac2113eb97392ddd6c6c> <http://data.lblod.info/vocabularies/erediensten/betrokkenBestuur> <http://data.lblod.info/id/betrokkenLokaleBesturen/0da11d9fdef2d45f90fd4ebd353ecd3c>.
+  }
+}
+
+;
+
+DELETE {
+  GRAPH ?g {
+      ?localInvolvementWortegemPetegem <http://data.lblod.info/vocabularies/erediensten/financieringspercentage> ?percentage.
+  }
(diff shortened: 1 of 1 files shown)


config/migrations/2025/20250107101243-delete-betrokken.sparql
```diff
@@ -0,0 +1,23 @@
...
+ }
+}
+insert {
+ graph ?g {
+    <http://data.lblod.info/id/betrokkenLokaleBesturen/0da11d9fdef2d45f90fd4ebd353ecd3c> <http://data.lblod.info/vocabularies/erediensten/financieringspercentage> 100.
+ }
+}
+where {
+ graph ?g {
+    <http://data.lblod.info/id/betrokkenLokaleBesturen/0da11d9fdef2d45f90fd4ebd353ecd3c> <http://data.lblod.info/vocabularies/erediensten/financieringspercentage> ?percentage.
```
mirdono, line 21: **Nitpick:** Personally, I tend to avoid repeating URIs as done in this update query. In this case I would use variable and a `VALUES` or `FILTER` statement to limit it the correct URI. Imo this makes queries a bit easier to read, given the use a meaningful variable name. You can, for example, add something like
```sparql
VALUES ?localInvolvementOudenaarde {
  <http://data.lblod.info/id/betrokkenLokaleBesturen/0da11d9fdef2d45f90fd4ebd353ecd3c>
}
```
and replace the repeated URIs with `?localInvolvementOudenaarde`.

config/migrations/2025/20250107101243-delete-betrokken.sparql
```diff
@@ -0,0 +1,23 @@
+delete where {
+ graph ?g {
```
mirdono, line 2: **Nitpick:** I tend to explicitly specify the graph when possible (and it does not overly complicate the query) to limit the chance for unintentional changes in other graphs, for example the landing zone graphs used by the consumers. (For this specific migration it should not matter as the modified data is only located in the `http://mu.semte.ch/graphs/worship-service` graph.)

config/migrations/2025/20250107101243-delete-betrokken.sparql
wolfderechter: Good idea

wolfderechter:
@mirdono thank you for these useful suggestions, I'm learning new things :)

wolfderechter:
Done!

PR #499, opened 2025-01-16
Hotfix/betuursfuncties fusiegemeentes

cecemel:
DL-6377: Attach the same instance of bestuursfuncties to Gemeente and OCMW.
Note; split the migration in 2 steps, seemed like an heavy query.


mirdono (changes requested):
Minor comment on the changelog. Migration itself looks good and works as expected.

PR #501, opened 2025-01-17
fix: merger change event between St.-Amandus and Sint-Martinus

mirdono:
The worship service the worship organisation *St.-Amandus* was not included as `orginalOrganization` in its merger change event. As a result it was not listed as involved organisation in the change event details. Furthermore, its status was not correctly updated to "Not active".

The new migration corrects the relevant change event by adding a link to the worship organisation and updating the organisation's status.

## How to test
- Restart the correct migrations service, see command in the changelog.
- Verify that
  + The status of *St.-Amandus* has changed to "Niet actief"
  + The merger change event should now list *St.-Amandus* as involved organisation.

The URIs of the relevant resources can be found in the ticket and the migration itself.

## Notes
- Should be tested with PROD data
- The index updates may take some time, depending on the available resources, the status of *St.-Amandus* in the organisations overview table will only change when they are done. The status on its core data page is immedialtly accurate (barring when cached data is used).
- On the core data page for *St.-Amandus* the date in the "Gewijzigd op ..." field below the status pill will **not** change. This date is taken from the most recent change event for which the organisation is a *resulting organisation* instead of considering all change events for the organisation. This will be improved in a separate ticket, the comments for the related ticket contain more background information.

## Related tickets
- OP-3534

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
 # Changelog
+## 1.28.5 (TODO)
+### Backend
+- Datafix: correct merger change events for worship organisations (OP-3534)
+### Deploy notes
+- `drc restart migrations-triggering-indexing; drc logs -ft --tail=200 migrations-triggering-indexing`
+
 ## 1.28.4 (2025-01-16)
 ### Backend
 #### Data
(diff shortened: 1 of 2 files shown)


claire-lovisa (approved):
all good :)

PR #507, opened 2025-01-23
Bump mandatarissen consumer to v0.1.0

claire-lovisa:
OP-3510

Bump the delta-consumer version of mandatarissen-consumer to v0.1.0, will all required changes in configuration and data flushing.


mirdono (commented):
Tested following the deploy instructions in the changelog. Works correctly as far as I can tell. Played around with the frontend to see mandatarissen-related functionality still works, encountered no (unknown) errors.

cecemel (approved):
Thanks, seems like working.

PR #508, opened 2025-01-24
remove province and commune classifications from admin graph

bdevloed:
Remove the province and commune classifications from the admin unit graph - as they are already stored in the shared graph.


mirdono (commented):
The migration itself looks good, two small comments about the PR itself:

**Suggestion:** Could you add an entry to the Unreleased section of the [changelog](https://github.com/lblod/app-organization-portal/blob/development/CHANGELOG.md). This makes it easier to keep track of things.

**Nitpick:** Could you also mention the ticket number in the PR's description or title?

bdevloed:
> The migration itself looks good, two small comments about the PR itself:
> 
> **Suggestion:** Could you add an entry to the Unreleased section of the [changelog](https://github.com/lblod/app-organization-portal/blob/development/CHANGELOG.md). This makes it easier to keep track of things.
> 
> **Nitpick:** Could you also mention the ticket number in the PR's description or title?

Thx for the suggestions. I've updated the description and updated the changelog.

PR #509, opened 2025-01-24
refactor: clean up search indexing configurations

mirdono:
This PR proposes to simplify the configuration for the search service.

### Remove the dev configuration
The use of two separate configurations adds some complexity. Every change must be applied to both configurations, although, such changes are fairly rare. This increases the risk of mistakes and/or the configuration diverging from each other. Not that there already were small differences between the two configurations. In my opinion this added complexity outweighs the speed gains one gets from using a dev-specific configuration. Especially since performing a full re-index tends to take an order of hours (for production data).

### Remove unused entries
All positions contained seemingly double properties `id` and `uuid`, I removed the latter one. Based on some testing this does not seem break anything. I might overlook something, as the frontend does do [something](https://github.com/lblod/frontend-organization-portal/blob/b457f29fe8b7af5c0c170e68ce22dc28059632f3/app/routes/people/index.js#L58) with the `uuid`. As far as I tested the entries processed in the frontend still have a `uuid` field, irrelevant of whether the index is present.

The indexes for the `organization_municipality` and `organization_province` are currently not used. Possibly they were added provisionally with functionality in mind which ended up not being added.

Removing this unneeded entries decreases the time needed for a full re-index. On my own machine this now takes less than an hour for DEV data. Your experience may vary depending on you machine and which data you use.

### Miscellaneous changes
Changed to ordering of entries so that
- they are in the same order for each type; and
- the entries in the properties and mappings sections are in the same order.

## Points for discussion
- The indexes for `first_name_used` are currently not used in the frontend, it only searches for `given_name`. We can either remove this entry from indexing or extend the filter in the frontend to also search both names.

## Notes
- I tried to keep the performed changes in separate commits to simplify finding the root cause if any bugs pop up.
- When using DEV data you may run into errors when trying to navigate to the details for some persons. This is unrelated to these changes. The persons in question lack a type triple for some reason, causing the frontend to crash.


claire-lovisa (commented):
The changes make sense. When testing though, even if it seems to work fine, I encountered some errors in the console. Can you check if they are related to this or something else? 
I tested on QA data, master branch for the frontend, and I found these:
- filtering by `gemeente` or `province` give 500 on `GET
	http://localhost:4200/organizations?filter[sub-organizations][:exact:name]=Aalst&filter[classification][id]=5ab0e9b8a3b2ca7c5e000000` I also had it for other municipalities/provinces, for worship and non-worship organizations
- on non-worship positions: sorting by organization or position gives error `While generating link to route "people.person": You must provide param `id` to `generate`.`, not sure if it's a data issue or a search issue though. Same error when filtering with organization "Gemeente Aalst"

mirdono:
Thanks for the additional testing and feedback!

> The changes make sense. When testing though, even if it seems to work fine, I encountered some errors in the console. Can you check if they are related to this or something else? I tested on QA data, master branch for the frontend, and I found these:
> 
>     * filtering by `gemeente` or `province` give 500 on `GET http://localhost:4200/organizations?filter[sub-organizations][:exact:name]=Aalst&filter[classification][id]=5ab0e9b8a3b2ca7c5e000000` I also had it for other municipalities/provinces, for worship and non-worship organizations

I am having trouble replicating the behaviour you described. For me the shown `GET` request succeeds with a 200. What I did:
1. Switch to master branches for backend and frontend
2. Load QA data in the backend
3. Use the search configuration from this PR
4. Perform a full re-index using the `reset-elastic.sh` script
5. Load up a local version of the frontend with`eds --proxy http://host`
6. Play around with the municipality and province filters on the organizations page (both worship and non-worship)

Did you do something different/extra/...?

> 
>     * on non-worship positions: sorting by organization or position gives error `While generating link to route "people.person": You must provide param `id`to`generate`.`, not sure if it's a data issue or a search issue though. Same error when filtering with organization "Gemeente Aalst"

This seems to be a data issue, the same happens on all three environments your might have to go through several pages before encountering it.) I could have sworn there was a ticket for this issue, but can't find it in Jira atm.

claire-lovisa (approved):
Indeed sorry for the false alarm, I was not using the correct pair of frnontend/backend branches. As far as I tested, it works like a charm :)

PR #512, opened 2025-02-05
Add dates mapping rules for mandatarissen and leidinggevenden consumers

claire-lovisa:
**OP-3544**

The mandatarissen and leidinggevenden consumers were missing a date formatting mapping that is making sure that the data we display to the user is of the same date as the one shown in the original app. This transcription was made by the reasoner and we forgot to include it in the mapping configuration.

I also updated the timestamp of migrations that need to run again when I'll re-deploy to QA and DEV. It will have no impact on the upcoming PROD deploy.


claire-lovisa:
@mirdono Indeed, those are properties of bestuursorganen (that are exported by the producer, I think it's why it has been put in the reasoner. But OP is the master of bestuursorganen so in this consumer we don't ingest those dates. This is why I didn't add them to the mapping rules.

PR #513, opened 2025-02-06
Remove reasoner service, replaced by new consumer logic

claire-lovisa:
After the consumer bumps done in #503, #506 and #507, we don't need the reasoner anymore. This PR removes it from the services and from the configuration.


mirdono (approved):
**Suggestion:** Could you add an entry to the changelog to bring the `reasoner`
service down *before* switching to the new tagged version? Otherwise I am likely
to forget that if this becomes part of a (pre-)release I would happen to deploy.

Otherwise, it looks good.

claire-lovisa:
Sorry, I forgot all the Changelogs haha. Will update it, thanks for noticing!

PR #514, opened 2025-02-07
chore: additional example environment settings for worship consumers

mirdono:
This adds example entries for `DCR_SYNC_LOGIN_ENDPOINT` and `DCR_SECRET_KEY`
environment settings for the worship consumers. The example value for
`DCR_SYNC_LOGIN_ENDPOINT` was chosen to match the value for `DCR_SYNC_BASE_URL`.

These environment settings need to be set correctly for both
consumers. Including examples in the `docker-compose.yml` makes this more
explicit.


claire-lovisa (approved):
Thanks!

PR #516, opened 2025-02-14
DO NOT MERGE: cleanup site data for DEV environment

mirdono:
:warning: Do **not** merge this PR. It is meant to simplify reviewing and testing the proposed migrations. The migrations will be executed locally only on DEV.

The data concerning organisation's sites got messed up at some point in the past. Most likely connected to a bug in the `sync-wegwijs-organization-service` which caused numerous issues with address data. (See [#8](https://github.com/lblod/sync-wegwijs-organization-service/pull/8) and the associated ticket for some background.) The migrations in this PR clean up at least some of the data.

I "clicked around" a bit in the frontend and everything seems to remain working. A second look would be appreciated.

## Remove organisation data from incorrect graphs
For several worship organisations some data was duplicated into the administrative unit graph. Same for non-worship organisations were data was duplicated in the worship graph. The affected data seems to be limited the organisation's UUID, type, and sites, along with their address and contact data. The first two migrations remove the duplicate data from the wrong graph.

I split this into two migrations, each removing from one graph, to keep the queries simpler and reduce the chance of deleting too much/the wrong data.

## Regenerate full addresses
Due to the bug mentioned above, many site addresses have multiple `locn:fullAddress` triples, often containing an unrelated or empty address string. To this end the migration first removes **all** `locn:fullAddress` triples from the worship and administrative-unit graphs. Afterwards it re-inserts new triples for all existing `locn:Address` resources in the same graphs.

The chosen approach of removing all full addresses only to re-insert most of them is a bit aggressive. But does allow to keep the queries rather straightforward compared to trying to delete only the incorrect triples etc.

## Remove site data from the public graph
A bunch of sites, along with their address and contact point data, were put into the public graph. This kind of data is not supposed to be in that graph. The last migration removes it all.

### DIFF
--- a/config/migrations/2025/20250213153859-dev-data-cleanup/20250213153859-delete-non-worship-organisations-from-worship-graph.sparql
+++ b/config/migrations/2025/20250213153859-dev-data-cleanup/20250213153859-delete-non-worship-organisations-from-worship-graph.sparql
@@ -1,0 +1,44 @@
+PREFIX organisatie: <https://data.vlaanderen.be/ns/organisatie#>
+PREFIX org: <http://www.w3.org/ns/org#>
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+
+# Remove triples for non-worship organisations from the worship graph. The
+# duplicated triples are limited to type and uuid predicates as well as site
+# data.
+DELETE {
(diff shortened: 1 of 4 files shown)


claire-lovisa (approved):
I also clicked around, checked the data and did a reindex: didn't see anything suspicious!

PR #517, opened 2025-02-17
Restore betrokken lokale besturen

claire-lovisa:
`ere:betrokkenBestuur` triples have accidentally been deleted in https://github.com/lblod/app-organization-portal/pull/515 (file `config/migrations/2025/20250213162849-update-worship-consumer-again/20250213162927-remove-no-types.sparql` because the type of the resource was in a different graph than this triple).

This migration restores them and should fix OP-3546


mirdono (approved):
Looks good, thanks for the quick fix.

PR #518, opened 2025-02-20
Hotfix/vgc classificatie code to public

cecemel:
Bestuursorgaanclassificatiecode of VGC was in wrong graph, so it wouldn't export for deltas producers. [DL-6428]


aatauil (approved):
Tested, 6 classification triples now in public graph

PR #519, opened 2025-02-21
chore: move status for municipalities and provinces in shared graph

mirdono:
For new frontend functionality concerning local involvements a worship user
should be able to read the status for municipalities and provinces. This
migration moves these triples to the `shared` graph such worship users have read
access.

The status for municipalities and provinces cannot be changed by users so they
do not require write access. The status of these types of organisations can only
be changed manually via migrations, for example when municipalities merge as was
done in #482. This PR makes the situation for statuses essentially the same as
was already the case for the (legal) names for municipalities and provinces.

This should not impact:
- The produced data as the producers already take the `shared` graph into
  account.
- The search indexes as this is based on the access rights.
- The consumers as OP is the master of this data.

## Related PRs
- Frontend PR for which this data move is needed: https://github.com/lblod/frontend-organization-portal/pull/670

## Related tickets
- OP-3469

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 ## Unreleased
+### Frontend
+- Bump to version [TODO](Link to frontend release)
+### Backend
+- Datafix: move status for municipalities and provinces to the `shared` graph such that it can be read by worship users [OP-3469]
+### Deploy notes
+```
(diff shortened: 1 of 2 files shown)


Riadabd (approved):

mirdono:
> Forgot to mention to update the CHANGELOG before merging, unless you were already planning to do so after the PR approval.

Thanks for the reminder, I had forgotten that.

PR #520, opened 2025-02-25
LDES feed for OP data

Rahien:
# Setup
You need to start the initial load of the LDES stream, do this by setting the environment var `WRITE_INITIAL_STATE` to true.
Once that is complete, your LDES files should be completely loaded and the files should appear in your data folder. You should now disable this environment variable again so the load isn't repeated.

# Checking if everything works
After the initial load, verify that
- an update of a meaningful concept (e.g. Bestuursorgaan, Mandaat) receives a dct:modified of the current datetime 
- after such an update, the current version of that instance is put on the LDES with its own versioned uri referring to the real instance uri (you can do this by just examining the contents of the file)
- if you're really want to go next level, you can verify that all concepts in the db are indeed present in the LDES (and vice versa) by running the auto-healing. This should find no missing instances or tombstones.

# Consuming this LDES feed
LMB has consumed this LDES feed in this branch: https://github.com/lblod/app-lokaal-mandatenbeheer/pull/391


JonasVanHoof:
Seems to be good ! 
```bash
# logs of the OP stack
ldes-delta-pusher-1  | Headers set on SPARQL client: {"requestDefaults":{"headers":{"mu-auth-sudo":"true"}}}
ldes-delta-pusher-1  | Found 0 missing values: []
ldes-delta-pusher-1  | No differences found.
ldes-delta-pusher-1  | Healing the LDES took 177460.78657599998 ms
```

claire-lovisa (approved):
Hi! I have been able to test the LDES stream following your instructions and it worked just as advertised. I also read the code and it makes sense, but I'm not so familiar with LDES so I might not be the best bug-spotter on this :sweat_smile:

cecemel (commented):
I have some concerns:

It seems the op-public delta producer provides the information you need. It would have been less work for everyone to just start consuming that.

We're introducing yet another serialization format with its own configuration; not only another way of configuring things but probably managing data that's maybe different in subtle ways. (i.e., oh, this producer misses type X, oh, and this has predicate Y).

Each of these services has its intricacies on an operational level. There will probably be bugs, getting the healing running, whatever corrective operations need to be done because something went wrong. We already have this for deltas, and now we have new things to deal with...

So, don't get me wrong, I am not arguing about the superiority of the format, the implementation, the developer-friendliness, (yes, we know the deal with deltas) or whatever. (It looks nice, by the way, and I like some of the ideas.)

My concerns are related to adding heterogeneity where it's not strictly needed, not taking the history and current state into account, the potential burden over the long run, and the fact that it's introduced without a plan.

Furthermore, I don't agree with publishing all ?p linked to a subject. This makes it easier to have accidents; i.e., publishing data that shouldn't be published, and we run the risk that consumers will start to rely on internal OP data needed to run the app. But that's an implementation detail.

PS: I am also aware that there was already an LDES feed available, but it was an unused stub, so I learned.

mirdono (commented):
In general it seems to work as advertised. Do have some questions/remarks though.

**Question:** While playing around with editing resources I noticed that sometimes multiple `dcterms:modified` triples are inserted for the same resource. For example, when I edit core data for an organisation (more specifically the alternative names**, it seems to insert one timestamp per inserted/deleted/updated triple. The timestamps are typically within one second, differing only a few 10 milliseconds. Subsequent edits tend to update all of the timestamps, but sometimes replace them with a single one. I am not sure whether this is the intended/expected behaviour. It seems at least odd to me.

**Remark/Question:** I agree with @cecemel that publishing all properties for a resource, which I presume is more than LMB uses, is not the best thing to do. It also includes data on worship services, which I do not think LMB needs. (Correct me, if I am wrong.) Is it feasible to limit the data produced by this producer? Would that significantly increase the complexity? (Maybe this is just a more a broader discussion concerning general producers vs. specific producers.)

**Opinion:** I understand @cecemel's concerns with respect to adding extra heterogeneity to OP. I am not sure that, on OP's side we currently have enough OP+LDES expertise to fix all issues that might pop. At least not without support from LMB. (Apologies if I underestimate anyone here. I mostly look at it from my own perspective.)

Rahien:
> **Question:** While playing around with editing resources I noticed that sometimes multiple `dcterms:modified` triples are inserted for the same resource. For example, when I edit core data for an organisation (more specifically the alternative names**, it seems to insert one timestamp per inserted/deleted/updated triple. The timestamps are typically within one second, differing only a few 10 milliseconds. Subsequent edits tend to update all of the timestamps, but sometimes replace them with a single one. I am not sure whether this is the intended/expected behaviour. It seems at least odd to me.

This is odd. i will investigate.

> **Remark/Question:** I agree with @cecemel that publishing all properties for a resource, which I presume is more than LMB uses, is not the best thing to do. It also includes data on worship services, which I do not think LMB needs. (Correct me, if I am wrong.) Is it feasible to limit the data produced by this producer? Would that significantly increase the complexity? (Maybe this is just a more a broader discussion concerning general producers vs. specific producers.)

We shouldn't set up an LDES feed per consumer of it, but rather an LDES feed as a statement 'we are publishing all of this data, here you can track when it changes and pick the bits that you are interested in yourself'
As you don't have any private information in OP (yet?) filtering predicates doesn't feel necessary (yet?)

> **Opinion:** I understand @cecemel's concerns with respect to adding extra heterogeneity to OP. I am not sure that, on OP's side we currently have enough OP+LDES expertise to fix all issues that might pop. At least not without support from LMB. (Apologies if I underestimate anyone here. I mostly look at it from my own perspective.)

While this is understandable, I think we should look at it from another perspective: if we are going to publish LDES feeds to our external clients, we should also use them internally (eat your own dogfood)

Rahien:
The reason for the duplicate dct:modifieds seems to be that you are using a custom delta-notifier that is pretty old and doesn't support merging of deltas (which the modified service relies upon). I used this version and it works:
`image: semtech/mu-delta-notifier:0.3.0`

mirdono:
> The reason for the duplicate dct:modifieds seems to be that you are using a custom delta-notifier that is pretty old and doesn't support merging of deltas (which the modified service relies upon). I used this version and it works:
> `image: semtech/mu-delta-notifier:0.3.0`

Thanks for investigating this and also finding a solution. I will look into bumping the used `mu-delta-notifier` to something more recent.

Rahien:
closed in favor of working with delta producers/consumers

PR #523, opened 2025-03-06
Add missing codes to public producer

claire-lovisa:
Some codes that should be produced by OP were missing in the public producer. We need it to complete the cleanup up the mandatarissen and leidinggevenden producers in Loket, hence the horfix.

To test it, you can start from a production copy, run a healing of the public producer
```
drc exec delta-producer-background-jobs-initiator curl -X POST http://localhost/public/healing-jobs 
```
Then checkout this branch and
```
drc restart migrations delta-producer-publication-graph-maintainer
drc exec delta-producer-background-jobs-initiator curl -X POST http://localhost/public/healing-jobs 
```

The missing codes should be produced by the healing

### DIFF
--- a/config/delta-producer/public/export.json
+++ b/config/delta-producer/public/export.json
@@ -497,6 +497,58 @@
         "http://mu.semte.ch/vocabularies/core/uuid",
         "http://www.w3.org/2004/02/skos/core#prefLabel"
       ]
+    },
+    {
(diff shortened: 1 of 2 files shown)


mirdono (approved):
Following the instructions I see the statuses appearing in the graph and in the generated delta (files). Looks correct to me.

PR #525, opened 2025-03-12
[OP-3103] Set up the jobs dashboard

Windvis:
Based on the same setup in the worship decisions app: https://github.com/lblod/app-worship-decisions-database/pull/29

It's not 100% working yet. The login works, but the errors page doesn't load (needs more mu-cl-resources configs, which the other PR didn't seem to include or already was part of the app). The jobs page does load, but there is no data. I have to double check if there is actual data in the DB or if something is wrong in the config.


README.md
```diff
@@ -185,3 +185,24 @@ To make sure the app can share data, producers need to be set up. There is an in
...
+To add a user, make sure to have installed [mu-cli](https://github.com/mu-semtech/mu-cli) first.
+Then in `docker-compose.override.yml`
+```
+  dashboard-login
+    environment:
+      MU_APPLICATION_SALT: 'a_random_string_with_sufficient_entropy_hence_not_this_one'
+```
+You can generate by running `mu script project-scripts generate-dashboard-login` and following the steps.
+Restart `migrations` and it should work.
+Note: on DEV and QA, the passwords will be kept in on the server in `docker-compose.override.yml`
```
mirdono, line 208: **Question:** Could you clarify this last sentence? I am not sure what you mean by it.

README.md
```diff
@@ -185,3 +185,24 @@ To make sure the app can share data, producers need to be set up. There is an in
...
+```
+127.0.0.1 dashboard.localhost
+```
+### login (all environments)
+For now, we use specific logins for the dashboard users. Each environment has its own passwords.
+
+To add a user, make sure to have installed [mu-cli](https://github.com/mu-semtech/mu-cli) first.
+Then in `docker-compose.override.yml`
+```
+  dashboard-login
```
mirdono, line 202: **Nitpick:** There is a colon missing after `dashboard-login`.

```suggestion
dashboard-login:
```

README.md
```diff
@@ -185,3 +185,24 @@ To make sure the app can share data, producers need to be set up. There is an in
        )
      }
 ```
+
+## configuring the dashboard
+### accessing the dashboard from your local machine
+Since we use dispatcher v2, which dispatches on hostname, we'll have to update `/etc/config/hosts`.
```
mirdono, line 191: **Nitpick:** Not sure how system-specific this is, but on my system the file is located at `/etc/hosts`.

config/authorization/config.ex
```diff
@@ -206,6 +219,37 @@ defmodule Acl.UserGroups.Config do
...
+                "http://lblod.data.gift/vocabularies/reporting/Report",
+                "http://open-services.net/ns/core#Error",
+                  "http://www.semanticdesktop.org/ontologies/2007/03/22/nfo#DataContainer",
+                  "http://www.semanticdesktop.org/ontologies/2007/03/22/nfo#FileDataObject",
+                  "http://www.semanticdesktop.org/ontologies/2007/03/22/nfo#DataContainer"
+              ]
+            }
+          },
+          %GraphSpec{
+            graph: "http://mu.semte.ch/graphs/system/jobs",
```
mirdono, line 241: **Issue:** Unfortunately OP also has a second graph in which it stores jobs: `<http://mu.semte.ch/graphs/jobs>`. To see all jobs the `dashboard-users` should have access to this graph as well. (The better solution would be to merge the graphs, but I would argue that is out of the scope for this ticket.)

config/resources/file.json
```diff
@@ -0,0 +1,107 @@
...
+                    "cardinality": "one",
+                    "inverse": true
+                }
+            },
+            "features": ["include-uri"],
+            "new-resource-base": "http://data.lblod.info/files/"
+        },
+        "remote-data-objects": {
+            "name": "remote-data-object",
+            "class": "nfo:RemoteDataObject",
```
mirdono, line 58: **Question:** What is this exactly used for? As far as I see OP currently has no resources of this kind.

config/resources/harvest.json
```diff
@@ -0,0 +1,40 @@
+{
+    "version": "0.1",
+    "prefixes": {
+        "adms": "http://www.w3.org/ns/adms#",
+        "dct": "http://purl.org/dc/terms/",
+        "hrvst": "http://lblod.data.gift/vocabularies/harvesting/",
+        "task": "http://redpencil.data.gift/vocabularies/tasks/"
+    },
+    "resources": {
+        "harvesting-collections": {
```
mirdono, line 10: **Question:** These kind of resources seem to be related to harvesters, which is not something OP (currently) uses. I am not sure whether it is worthwhile to add configurations for stuff we are not using.

docker-compose.yml
```diff
@@ -147,6 +147,14 @@ services:
     logging: *default-logging
     links:
       - db:database
+  dashboard-login:
+    image: semtech/mu-login-service:3.0.0
+    links:
+      - triplestore:database
```
mirdono, line 153: **Issue:** Is there are reason to go directly to the `triplestore` instead of via the mu-authorization service, called `db` here? The example PR you linked does go via the mu-authorization service.

README.md
Windvis: I'm not sure, I copied that as-is from the worship DB PR. @cecemel Any insights?

README.md
Windvis: Yea, I assume that was a mistake in the docs I copied. I didn't take a detailed look when I copied it :smile:

config/authorization/config.ex
Windvis: Ok, I'll add it :+1:

config/resources/file.json
Windvis: I would have to double check. There were some configs I needed to add to make the frontend-dashboard app happy, otherwise mu-cl-resources would return a 500. It is possible I added too much config though.

docker-compose.yml
Windvis: No specific reason. I think I changed that when debugging an issue but it wasn't the real cause.

I don't think it really matters though, since the service uses sudo queries AFAIK, but I'll change it back :+1:

README.md
Windvis: I just removed it :smile:

mirdono (changes requested):
**Issue:** Could you also update the changelog and add the appropriate deploy instructions there?

config/resources/file.json
Windvis: There is a "remote data objects" tab on the task details page that tries to fetch that resource:

image

mirdono: Aah, I see. Clearly I did not click around far enough. Thanks for pointing this out.

Windvis:
> **Issue:** Could you also update the changelog and add the appropriate deploy instructions there?

Done :+1:

I also force pushed to clean up the git history a bit so a reset is needed if you want to test.

mirdono (approved):
Looks good. Thanks for figuring out how to set this up :pray:

Windvis:
Whoops, it seems I messed up the git history a bit and some of the changes in the second commit should have been in the first one, my bad.

PR #526, opened 2025-03-12
Add missing `logging` keys for some services

Riadabd:
## Ticket ID

DL-6490

## Ticket Description

Some services are missing docker compose keys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Backend
 - Datafix: move status for municipalities and provinces to the `shared` graph such that it can be read by worship users [OP-3469]
 - Bump `triplestore` and `publication-triplestore` to `v1.3.0-rc.1`. [OP-2492] [OP-3547]
+- Add missing docker compose keys. [DL-6490]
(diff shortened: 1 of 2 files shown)


claire-lovisa (commented):
I have a vague, very vague memory of us disabling the logs on ES and search but I can't find it back in the PRs so I might be wrong. @nbittich does it ring a bell maybe? Something about the logs taking up too much space..?

nbittich:
nope, I don't recall disabling logs  specifically for disk space reason 😬 
If that was the case we would probably set the label as `logging=false` , so the key is there but 'in theory' it doesn't do anything

claire-lovisa (approved):
then I approve :grin: thanks for your input @nbittich :)

Riadabd:
@claire-lovisa Just to be sure, the students are okay then as is? I have refrained from adding `labels: logging=true` for services that are missing it across different projects since it leads to extra disk usage, but I wanted to fully confirm first before merging.

claire-lovisa:
@Riadabd yes like that it's perfect! 👌

PR #528, opened 2025-03-20
Update the way we match the dashboard routes

Windvis:
The previous implementation only allowed the domain name to start with `dashboard.*`. However, we also want to prefix it with the environment name (for exapmle `dev.dashboard.*`) so our matchers needed to be updated.

To test, edit your `/etc/hosts` file and add both a `dev.dashboard.local` and `dashboard.local` entry to see if both correctly return the dashboard frontend (and logging in works).


Windvis:
No longer needed since we're changing the domains to `dashboard.dev.*`.

PR #490, opened 2025-01-07
Import 2025 NIS codes and update werkingsgebieden

claire-lovisa:
# Context

OP-3342

To follow up with the addition of mergers in OP, we want to add their NIS codes as well into the app. A (new version)[https://github.com/belgif/vocab-belgif/blob/master/prod/data/codelist/refnis2025.nt} of the NIS codes has been created to reflect the state of 2025.

This PR does a few things:
1. Import the 2025 NIS codes
2. Update our `werkingsgebieden` to link to 2025 NIS codes when an exact match is found between the version of 2019 and the one of 2025
3. Add a link from `werkingsgebieden` to 2025 NIS codes for the mergers

# Expected output

The non-mergers should still look the same as before in the frontend but the data should be using mostly 2025 NIS codes.
The mergers should show the following NIS codes:
|                    Fuserende gemeenten                   |           Nieuwe gemeenten           |
|:--------------------------------------------------------:|:------------------------------------:|
|           Antwerpen (11002) + Borsbeek (11007)           |           Antwerpen (11002)          |
|    Galmaarden (23023) + Gooik (23024) + Herne (23032)    |          Pajottegem (23106)          |
|            Ruiselede (37012) + Wingene (37018)           |            Wingene (37021)           |
|             Meulebeke (37007) + Tielt (37015)            |             Tielt (37022)            |
|            De Pinte (44012) + Nazareth (44048)           |       Nazareth-De Pinte (44086)      |
|          Lochristi (44034) + Wachtebeke (44073)          |           Lochristi (44087)          |
|            Lokeren (46014) + Moerbeke (44045)            |            Lokeren (46029)           |
|             Melle (44040) + Merelbeke (44043)            |        Merelbeke-Melle (44088)       |
| Beveren (46003) + Kruibeke (46013) + Zwijndrecht (11056) | Beveren-Kruibeke-Zwijndrecht (46030) |
|             Bilzen (73006) + Hoeselt (73032)             |        Bilzen-Hoeselt (73110)        |
|            Borgloon (73009) + Tongeren (73083)           |       Tongeren-Borgloon (73111)      |
|             Ham (71069) + Tessenderlo (71057)            |        Tessenderlo-Ham (71071)       |
|            Hasselt (71022) + Kortessem (73040)           |            Hasselt (71072)           |

PR #533, opened 2025-03-25
Produce NIS codes and countries in public producer

claire-lovisa:
# Context

[CLBV-980]

We want to stop using the `organizations-public-info` gradually. For that, in `app-contactgegevens`, we need the public producer to produce the NIS codes (which are public so it fits with the description :) ). This is what this PR does.

**_Update_**: we also need the countries there, it's a codelist and OP is the master of it so it makes sense to also produce it I think.

# Test instructions

On a running stack:
```
drc restart migrations delta-producer-publication-graph-maintainer 
drc exec delta-producer-background-jobs-initiator curl -X POST http://localhost/public/healing-jobs
```
And check that the produced deltas are containing the NIS codes and the countries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,14 @@
 - Set up the dashboard app [OP-3103]
 - Datafix: correct date of change event of kerkfabriek st petrus [OP-3555]
 - Update OCMWv to VVMW [OP-3565] and back. It's technically a nil-operation. But you should run the migrations nevertheless.
+- Extend the public producer to include NIS codes [CLBV-980]
 ### Deploy notes
 ```
 drc restart migrations-triggering-indexing
 drc restart migrations; drc logs -ft --tail=200 migrations
 drc pull frontend; drc up -d frontend
 drc up -d mandatarissen-consumer leidinggevenden-consumer worship-services-main-info-consumer worship-services-private-info-consumer
+drc restart delta-producer-publication-graph-maintainer
 ```
 #### For upgrade databases
 [This README](https://github.com/Riadabd/upgrade-virtuoso) provides the necessary steps for upgrading the database. **NOTE**: This will involve shutting down the app for small period of time (around 30 minutes).
--- a/config/delta-producer/public/export.json
+++ b/config/delta-producer/public/export.json
@@ -564,6 +564,31 @@
         "http://www.opengis.net/ont/geosparql#sfWithin",
         "http://www.w3.org/2004/02/skos/core#exactMatch"
       ]
+    },
+    {
+      "type": "http://www.w3.org/2004/02/skos/core#Concept",
+      "graphsFilter": [
+        "http://mu.semte.ch/graphs/public"
+      ],
+      "pathToConceptScheme": [
+        "http://www.w3.org/2004/02/skos/core#inScheme"
+      ],
+      "properties": [
+        "http://mu.semte.ch/vocabularies/core/uuid",
+        "http://www.w3.org/2004/02/skos/core#altLabel",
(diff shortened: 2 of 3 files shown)


mirdono (changes requested):
**Issue:** I am not sure we should keep the NIS2019 data around. It was superseded by the NIS2025 version. Furthermore, the data does not seem to be actively used anymore, at least not since the migrations in #490 were executed. Unless you think otherwise, I would avoid producing the NIS2019 data so that we can more easily remove it altogether.

mirdono:
> @mirdono You do have a point, but I'm hesitant because we have occurrences of active admin units linked to a location for which the NIS code only seems to exist in the 2019 realm. It might mean there is a problem in the data, but it feels a bit less scary to produce a little too much rather than a little to few. Also it gives kind of an history on the different locations, plus it's not that much data so it's not causing extra load. What do you think?
> 
> An example is `http://data.lblod.info/id/bestuurseenheden/fc4ab3a9032d05f4f9adcee44f0186a552fbd8c769cd531a633bef83d0d1e631` which is active and is in the location `http://data.lblod.info/id/werkingsgebieden/4257b4b87f9bc4b6ee50ddb606d473cbdf274eed9c5e3e4edb43d3cf5fffd68c` with for exact NIS match `http://vocab.belgif.be/auth/refnis2019/46013`

Apologies, my mistake here. I was confusing the NIS codes and NUTS/LAU codes, which are related but not the same. They used different update strategies concerning the latest municipality mergers. In NUTS/LAU they ported the merging municipalities to the new version but set an end date for them, for instance your example still [exists](https://github.com/Informatievlaanderen/ipdc-lpdc/blob/cd39fad8d99f6f53a7b973876e66251a686a3ef9/codelijsten/geografisch-toepassingsgebied.ttl#L2335) in the new version. But in NIS only the merged municipality exists, which link back to the NIS2019 codes of the merging municipalities. So calling NIS2019 "superseded", as I did, is not correct.


I am wondering now why the `organizations-public-info` was producing this data in the first place. The PR #373 introducing it is not very informative. Should the `prov:Location` that #524 added to the public producer not be enough for the contactgegevens app? If it needs the whole NIS code list, why not just import it once? (Sorry, I should have thought of this before requesting to add this stuff to the producer here.)

claire-lovisa:
@mirdono I think the NIS codes were initially produced in `organizations-public-info` for app-contactgegevens-loket actually. As shows the doc [here](https://abb-vlaanderen.gitbook.io/handleiding-contact-en-organisatiegegevens/kerngegevens/overzicht-kerngegevens) they also have the "core data" tab where they want to show the NIS code amongst other things.

Why not importing it once, I think it would be the same debate for classification codes etc no? We could, but it's more maintenance when a change happens because then you need to propagate the migration to all the apps that are using it. If it's produced, the changes just propagate. If the NIS codes get again updated in 6 years for example, it's saving us some work in that distant future haha

mirdono:
> @mirdono I think the NIS codes were initially produced in `organizations-public-info` for app-contactgegevens-loket actually. As shows the doc [here](https://abb-vlaanderen.gitbook.io/handleiding-contact-en-organisatiegegevens/kerngegevens/overzicht-kerngegevens) they also have the "core data" tab where they want to show the NIS code amongst other things.
> 
> Why not importing it once, I think it would be the same debate for classification codes etc no? We could, but it's more maintenance when a change happens because then you need to propagate the migration to all the apps that are using it. If it's produced, the changes just propagate. If the NIS codes get again updated in 6 years for example, it's saving us some work in that distant future haha

You are probably right. The municipality mergers are already complicated enough, if we can automate some changes instead of having to do them manually I suppose it is worth it.

mirdono (approved):

PR #534, opened 2025-03-28
Add sites and status to representative organ in public producer

claire-lovisa:
The sites and status are actually produced because representative organs have  also the type `besluit:Bestuurseenheid`. But for the sake of consistency, adding them to `RepresentatiefOrgaan` makes sense.


mirdono (approved):
Looks good.

**Comment:** Representative bodies are actually **not** administrative units. In fact they only have the `besluit:Bestuurseenheid` type on DEV. (We might opt to remove it there as well for the sake of consistency.) So on the other environments this data is not yet produced (unless I am missing something else, I have not explicitly checked. :wink:)

claire-lovisa:
@mirdono aaaaah that's why data was produced on dev data but not locally on a prod copy! Makes sense now, thanks for the precision haha

PR #535, opened 2025-03-28
Remove change events from ckb's and update their status

wolfderechter:
## ID

OP-3562

## Description

Removal of change events of 8 ckb's + update of their status


config/migrations/2025/20250327135321-delete-change-events.sparql
```diff
@@ -0,0 +1,54 @@
+PREFIX core: <http://mu.semte.ch/vocabularies/core/>
+PREFIX org: <http://www.w3.org/ns/org#>
+PREFIX lblodorg: <http://lblod.data.gift/vocabularies/organisatie/>
+PREFIX euro: <http://data.europa.eu/m8g/>
+PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
+
+DELETE {
+  GRAPH <http://mu.semte.ch/graphs/worship-service> {
+    ?event ?ep ?eo ;
+           euro:hasFormalFramework ?decision .
```
mirdono, line 10: **Question:** Is this line necessary? These triples should already be covered by the previous line.

mirdono (changes requested):
**Issue:** Could you move the migration updating the organisation's statuses to the `migrations-triggering-indexing/2025` folder? This migrations service goes via the authorisation service, thereby automatically triggering a re-index the changed statuses. That way the status shown in the organisations overview table is automatically updated as well. (Otherwise, we would have to manually trigger a full re-index for just a handful of changed statuses.) Note, you will have to remove the comments in the migration, `mu-authorization` cannot properly deal with that. (This should change once we switch to the new `sparql-sparser` instead.)

**Issue:** Could you also update the changelog, so whoever deploys this does not forget to restart the correct services.

config/migrations/2025/20250327135321-delete-change-events.sparql
wolfderechter: Ah yes indeed, it's not needed

mirdono (approved):
Looks good now

PR #536, opened 2025-04-01
Produce secondary sites in public producer as well

claire-lovisa:
Sorry for yet again a PR on the producer, while testing I noticed I forgot to produce secondary sites as well :sweat_smile: 

I didn't add any extra changelog because it's fully included in the previous notes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 # Changelog
 ## Unreleased
 ### Backend
-- Extend the public producer to include NIS codes and more info in representative organs (sites, status) [CLBV-980]
+- Extend the public producer to include: [CLBV-980]
+  - NIS codes
+  - Sites and statuses of representative organs
+  - Secondary sites with their contacts and addresses
+  - Site descriptions
+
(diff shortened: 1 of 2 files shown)


mirdono (changes requested):
**Issue:** The `pathToConceptScheme`for `locn:Address` and `schema:ContactPoint` are currently defined only for `org:hasPrimarySite`. Assuming you also want to produce the actual addresses and contact points for secondary sites these will have to be extended.

mirdono (commented):
**Question:** I noticed that for sites `dcterms:description` triples are not produced. Is this intentional? This triple was added in #371 which, if I recall correctly, was also related to functionality in the contactdata app.

claire-lovisa:
@mirdono For the issue you're completely right, thanks for avoiding me one more PR :sweat_smile: I also added the path for secondary sites. And for the description, I was not aware that we had site descriptions and I missed it when I read CLB code but we want it for sure! I also added it to the PR.

mirdono (approved):

PR #537, opened 2025-04-09
Add migration and instructions to reset kbo sync

claire-lovisa:
# Context

[OP-3584]

To be tested with https://github.com/lblod/sync-wegwijs-organization-service/pull/9

Follow the changelog for testing instructions! The idea is to set the "modified" date of the KBO organizations extremely far in the past so that they will be updated again with the proper data types. Once the healing ran, you should be able to see statuses again in the frontend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,13 @@
 # Changelog
+## Unreleased
+### Backend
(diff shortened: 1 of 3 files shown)


mirdono (approved):
Looks good, don't forget to bump the service version before merging.

PR #538, opened 2025-04-17
fix: update NIS2019 end dates and link werkingsgebieden

mirdono:
The NIS2025 codes were initially introduced in #490. This PR resolves two small oversights in that initial import.

## End dates NIS2019 codes
Several NIS2019 were given an actual end date to indicate that they are no longer valid after the 2025 municipality merger. This PR updates the relevant end dates based on the most recent version of the NIS2019 code list. This way we know which NIS codes are no longer valid.

The used migration essentially replaces all of OP's NIS2019 code end dates that differ from those on the most recent version code list, and replaces them by the latter. Note that this covers more than just the NIS2019 affected by the 2025 municipality mergers.

## Update werkingsgebieden for Antwerp and East Flanders
As part of the 2025 municipality mergers, the (former) municipality of Zwijndrecht changed province from Antwerp to East Flanders. This change is reflected in the NIS codes by relating their NIS2019 and NIS2025 with `skos:narrowMatch` and `skos:broadMatch` instead of `skos:exactMatch`. Similarly, the municipality of Antwerp changed due to the addition of Borsbeek as a new district. More concretely:

- [East Flanders](https://github.com/lblod/app-organization-portal/blob/development/config/migrations/2025/20250107100000-nis-codes-2025/20250107101800-refnis2025.ttl#L7698): `<http://vocab.belgif.be/auth/refnis2025/40000> skos:narrowMatch <http://vocab.belgif.be/auth/refnis2019/40000>`
- [Antwerp (province)](https://github.com/lblod/app-organization-portal/blob/539376c47c6cc8aef21c3e88721b9ce65049950b/config/migrations/2025/20250107100000-nis-codes-2025/20250107101800-refnis2025.ttl#L7490): `<http://vocab.belgif.be/auth/refnis2025/10000> skos:broadMatch <http://vocab.belgif.be/auth/refnis2019/10000>`
- [Antwerp (municipality)](https://github.com/lblod/app-organization-portal/blob/development/config/migrations/2025/20250107100000-nis-codes-2025/20250107101800-refnis2025.ttl#L31): `<http://vocab.belgif.be/auth/refnis2025/11002> skos:narrowMatch <http://vocab.belgif.be/auth/refnis2019/11002>`

Since the initial migration relied on a `skos:exactMatch` relation between NIS codes, these three instances were missed and the linked werkingsgebieden were not updated.

Note, a similar situation occurs in other werkingsgebieden, such as Hasselt, but these are not (correctly) matched to their NIS code. This will be further investigated and fixed in follow-up PRs focusing on the werkingsgebieden data.

## Related tickets
- OP-3566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,14 @@
 ## Unreleased
 ### Backend
 - Fix KBO statuses not being displayed [OP-3584]
+- datafix: update NIS2019 end dates and link additional NIS2025 to werkingsgebieden [part of OP-3566]
 ### Deploy Notes
 ```
 drc restart migrations; drc logs -ft --tail=200 migrations
 drc up -d kbo-data-sync
 drc exec kbo-data-sync curl -X POST http://localhost/sync-all-kbo-data
 ```
+
 ## v1.31.2 (2025-04-02)
 ### Backend
 - Extend the public producer to include: [CLBV-980]
--- a/config/migrations/2025/20250416181230-fix--refnis-2025/20250416181230-refnis2019-after-2025-municipality-mergers.graph
+++ b/config/migrations/2025/20250416181230-fix--refnis-2025/20250416181230-refnis2019-after-2025-municipality-mergers.graph
@@ -1,0 +1,1 @@
+http://mu.semte.ch/graphs/refnis2019-data-temp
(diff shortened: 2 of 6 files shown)


claire-lovisa (approved):
Thanks for the cleanup!

PR #539, opened 2025-04-29
Hotfix/remove has suborganization from inactive ckb deinze

cecemel:
For context, see [DL-6604].

There is a 'double' instance of CKB Deinze in the database. One is active, and the other is inactive. This confuses systems, and it doesn't make sense that active organizations are linked (almost managed) by an inactive organization.

This commit fixes this issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,12 @@
 # Changelog
+## Unreleased
+### Backend
+- Extend the public producer to include NIS codes and more info in representative organs (sites, status) [CLBV-980]
+### Deploy notes
+```
+drc restart delta-producer-publication-graph-maintainer
+```
+
 ## v1.31.1 (2025-04-01)
 ### Backend
 - Removal of change events of 8 ckb's + update of their status [OP-3562]
@@ -7,6 +15,7 @@
 drc restart migrations-triggering-indexing; drc logs -ft --tail=200 migrations-triggering-indexing
 drc restart migrations; drc logs -ft --tail=200 migrations
 ```
+
 ## v1.31.0 (2025-03-27)
 ### Frontend
 - Bump to version [v1.30.0](https://github.com/lblod/frontend-organization-portal/blob/v1.30.0/CHANGELOG.md#v1300-2025-03-27)
@@ -18,14 +27,12 @@ drc restart migrations; drc logs -ft --tail=200 migrations
 - Set up the dashboard app [OP-3103]
 - Datafix: correct date of change event of kerkfabriek st petrus [OP-3555]
 - Update OCMWv to VVMW [OP-3565] and back. It's technically a nil-operation. But you should run the migrations nevertheless.
-- Extend the public producer to include NIS codes and more info in representative organs (sites, status) [CLBV-980]
 ### Deploy notes
 ```
 drc restart migrations-triggering-indexing
 drc restart migrations; drc logs -ft --tail=200 migrations
 drc pull frontend; drc up -d frontend
 drc up -d mandatarissen-consumer leidinggevenden-consumer worship-services-main-info-consumer worship-services-private-info-consumer
-drc restart delta-producer-publication-graph-maintainer
 ```
 #### For upgrade databases
 [This README](https://github.com/Riadabd/upgrade-virtuoso) provides the necessary steps for upgrading the database. **NOTE**: This will involve shutting down the app for small period of time (around 30 minutes).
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,12 @@
 # Changelog
+## v1.31.3 (2025-04-02)
```
mirdono, line 2: **Nitpick:** Set the correct release date.

I assume this is a copy-paste issue, but could you put the correct release date in the section title. Side note, the commit message of the second commit contains v1.31.**2** instead of v1.31.**3**.

config/migrations-triggering-indexing/2025/20250429103515-remove-has-suborganization-from-inactive-ckb-deinze.sparql
```diff
@@ -0,0 +1,31 @@
+INSERT {
+  GRAPH <http://mu.semte.ch/graphs/worship-service> {
+    <http://data.lblod.info/id/centraleBesturenVanDeEredienst/7f5475cfb202d12f54779f046441c9e1>
+      <http://www.w3.org/ns/org#hasSubOrganization> ?sub.
+  }
```
mirdono, line 5: **Issue:** You need to also insert the corresponding `org:membership` resources.

OP uses `org:membership`s to show/edit the related organisations in the frontend. These memberships should also be created here, otherwise these inserted relations will not show up for the users.

In this case inserting something like

``` sparql
  <http://data.lblod.info/id/lidmaatschap/UUID> a org:Membership ;
    mu:uuid UUID ;
    org:organization <http://data.lblod.info/id/centraleBesturenVanDeEredienst/7f5475cfb202d12f54779f046441c9e1> ;
    org:member ?sub ;
    org:role <http://data.lblod.info/id/rollen/4ec7d5c39bdc4e84b4174379b9e22ad8> .
```

along with each `org:hasSubOrganization` should work. (The role on the last line is the one for the generic "has a relation with" membership, as is used in this situation.)

You can find the relevant part of the domain model [here](https://github.com/lblod/app-organization-portal/blob/master/config/resources/domain.json#L834-L859).

For more background, I originally used [this migration](https://github.com/lblod/app-organization-portal/blob/master/config/migrations/2024/20240704112300-related-organizations-as-memberships/20240704112800-related-members.sparql#L1) to convert `org:hasSubOrganization` triples to memberships. There you can see the suborganisation was assigned as `org:member` in the membership and the, in this case, CKB as `org:organization`.

config/migrations-triggering-indexing/2025/20250429103515-remove-has-suborganization-from-inactive-ckb-deinze.sparql
```diff
@@ -0,0 +1,31 @@
...
+}
+
+;
+
+DELETE {
+  GRAPH <http://mu.semte.ch/graphs/worship-service> {
+    <http://data.lblod.info/id/centraleBesturenVanDeEredienst/643ddbd9448ca43828702527af7bfbb8>
+      <http://www.w3.org/ns/org#hasSubOrganization> ?sub.
+  }
```
mirdono, line 24: **Issue:** Also delete the corresponding `org:membership` resource.

Similar to the above comment with inserting the new relations. If you do not delete the corresponding `org:membership` resources, the `db-cleanup` service will recreate the `org:hasSubOrganization` triples due to [this job](https://github.com/lblod/app-organization-portal/blob/master/config/migrations/2024/20240930124819-membership-cleanup-jobs/20241001071404-membership-cleanup--insert-worship-has-sub-organization.ttl#L1).

In this case extending the `WHERE` clause with something like

``` sparql
  ?membership a org:Membership ;
    org:organization <http://data.lblod.info/id/centraleBesturenVanDeEredienst/643ddbd9448ca43828702527af7bfbb8> ;
    ?p ?o .
```

and than the `DELETE` clause with 

``` sparql
  ?membership ?p ?o .
```

should be enough.

Riadabd:
@mirdono I updated the migration based on the comments you provided. I ran it locally, and it looks like it's removing/inserting data correctly. Can you have a quick look at the changes? Once approved, I can squash the commits (to fix the typo in the second commit), tag a hotfix and merge the branch.

mirdono (changes requested):
One blocking issue remaining, after fixing that the migration indeed works as expected.
Also, don't forget to update the ate in the changelog entry once done.

PR #540, opened 2025-04-30
Fix: update change event date of ckb olen

wolfderechter:
## ID

OP-3594

## Description

Update date of a change event of ckb olen to '30-11-2023'.


mirdono (approved):
Looks good

PR #541, opened 2025-04-30
chore: move memberships from shared to administrative-unit graph

mirdono:
These memberships can be edited by users via the related organizations page. If
this data is kept in the, read-only, shared graph it leads to unexpected
behaviour when users (try to) remove such a membership. More specifically, as
the user has no write acces to the graph the request will be silently ignored
and the "removed" data will simply reappear after saving the data.

Furthermore, when a user adds similar memberships via the frontend they will be
added to the administrative-unit graph. Over time this can lead to an unexpected
and confusing distribution of memberships over graphs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - Fix KBO statuses not being displayed [OP-3584]
 - datafix: update NIS2019 end dates and link additional NIS2025 to werkingsgebieden [part of OP-3566]
 - datafix: correct date of change event of ckb olen [OP-3594]
+- datafix: move memberships from shared to administrative unit graph
 ### Deploy Notes
 ```
 drc restart migrations; drc logs -ft --tail=200 migrations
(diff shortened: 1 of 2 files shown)


claire-lovisa (approved):
Ohhh good catch! Out of curiosity, do you happen to know how we ended up in that situation?
In any case, tested to delete a relation before and after the migration and checked the data, seems all good.

mirdono:
> Ohhh good catch! Out of curiosity, do you happen to know how we ended up in that situation?
> In any case, tested to delete a relation before and after the migration and checked the data, seems all good.

I did not look into it in detail (as I wanted to finish the PRs before the long weekend). But I suspect the original triples (`org:hasSubOrganization`, `org:linkedTo`, ... ) from which these memberships were created were already spread across the graphs like this. The migrations converting them to memberships just inserted them into the same graph (see #454). Before the memberships these relations were not editable by so that was not an issue.

PR #542, opened 2025-04-30
chore: cleanup location data

mirdono:
This PR cleans up the existing location (*nl. werkingsgebieden*) data in preparation for adding functionality to view and edit them in OP. More specifically the migrations do the following:

1. Move all existing locations to the `public` graph. Previously the locations were largely duplicated in the `shared` and `adminsitrative-unit` graphs. I opted to put them in the `public` graph instead of in the `shared` graph as the existing locations are essentially a "standard" set that should not change over time. (Except in rather rare circumstances such as municipality mergers.)
2. Add the missing location for the district of Borsbeek. This was overlooked when creating the new district in #445
3. Add `geo:sfWithin` relations between the different location levels:
   a. Link all district locations to the location for the municipality of Antwerp
   b. Link all municipality locations the location of the province they are located in

## Notes
- When writing the last migration I noticed some issues concerning the placement of the memberships relating districts, municipalities, and provinces. For this migration to work correctly the migration proposed in #541 should be executed first. (Which is also why the target branch of the current PR is not development the branch from that PR.)

## Related ticket
- OP-3566

### DIFF
--- a/config/delta-producer/organizations/export.json
+++ b/config/delta-producer/organizations/export.json
@@ -389,7 +389,7 @@
     {
       "type": "http://www.w3.org/ns/prov#Location",
       "graphsFilter": [
-        "http://mu.semte.ch/graphs/shared"
+        "http://mu.semte.ch/graphs/public"
       ],
       "properties": [
         "http://www.w3.org/2000/01/rdf-schema#label",
--- a/config/delta-producer/public/export.json
+++ b/config/delta-producer/public/export.json
@@ -628,8 +628,7 @@
     {
       "type": "http://www.w3.org/ns/prov#Location",
       "graphsFilter": [
-        "http://mu.semte.ch/graphs/shared",
-        "http://mu.semte.ch/graphs/administrative-unit"
+        "http://mu.semte.ch/graphs/public"
       ],
       "hasRegexGraphsFilter": false,
       "properties": [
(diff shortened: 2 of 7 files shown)


config/migrations/2025/20250430130307-chore--cleanup-werkingsgebieden-data/20250430153558-feat--add-werkingsgbied-for-borsbeek-district.sparql
```diff
@@ -0,0 +1,24 @@
...
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
+
+INSERT {
+  GRAPH <http://mu.semte.ch/graphs/public> {
+    ?borsbeekLocation a prov:Location;
+                      mu:uuid ?uuid ;
+                      rdfs:label "Borsbeek" ;
+                      ext:werkingsgebiedNiveau "District" ;
+                      geo:sfWithin ?antwerpLocation .
```
claire-lovisa, line 12: You miss `?antwerpLocation` in your `SELECT`, causing that triple not to be inserted!

config/delta-producer/organizations/export.json
```diff
@@ -389,7 +389,8 @@
     {
       "type": "http://www.w3.org/ns/prov#Location",
       "graphsFilter": [
-        "http://mu.semte.ch/graphs/shared"
+        "http://mu.semte.ch/graphs/shared",
+        "http://mu.semte.ch/graphs/public"
```
claire-lovisa, line 393: We can only keep the public graph now that all the Locations are in there no? I checked and no location is left in the shared graph

config/delta-producer/public/export.json
```diff
@@ -629,7 +629,8 @@
       "type": "http://www.w3.org/ns/prov#Location",
       "graphsFilter": [
         "http://mu.semte.ch/graphs/shared",
-        "http://mu.semte.ch/graphs/administrative-unit"
+        "http://mu.semte.ch/graphs/administrative-unit",
+        "http://mu.semte.ch/graphs/public"
       ],
```
claire-lovisa, line 634: Same, only keep the public graph?

claire-lovisa (changes requested):

PR #546, opened 2025-05-21
fix: datafix of 3 orgs

wolfderechter:
## ID
OP-3602, OP-3604 and OP-3598

## Description

This PR bundles the datafix of 3 different orgs/tickets for PROD data:
- update status kerkfabriek weelde [OP-3598]
- delete change event and update status protestantse kerk oostende [OP-3604]
- update change event date islamic assocation [OP-3602]

## How to test

Setup stack with prod data, verify for each ticket that the desired change is done

### DIFF
--- a/config/migrations/2025/20250520105233-delete-change-event-oostende.sparql
+++ b/config/migrations/2025/20250520105233-delete-change-event-oostende.sparql
@@ -1,0 +1,32 @@
+PREFIX core: <http://mu.semte.ch/vocabularies/core/>
+PREFIX org: <http://www.w3.org/ns/org#>
+PREFIX lblodorg: <http://lblod.data.gift/vocabularies/organisatie/>
+PREFIX euro: <http://data.europa.eu/m8g/>
+PREFIX besluit: <http://data.vlaanderen.be/ns/besluit#>
+
+DELETE {
+  GRAPH <http://mu.semte.ch/graphs/worship-service> {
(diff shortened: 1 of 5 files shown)


config/migrations/2025/20250520105233-delete-change-event-oostende.sparql
```diff
@@ -0,0 +1,32 @@
...
+  GRAPH <http://mu.semte.ch/graphs/worship-service> {
+    ?event a org:ChangeEvent ;
+           core:uuid "6824418BD4DB18EC8BA95940" ;
+           lblodorg:veranderingsgebeurtenisResultaat ?result ;
+           ?ep ?eo .
+
+    ?result a lblodorg:VeranderingsgebeurtenisResultaat ;
+            ?rp ?ro .
+
+    OPTIONAL {
```
mirdono, line 26: **Comment (non-blocking):** Superfluous clause as change event has no linked decision.

The change event in question is not linked to a decision so strictly speaking this part is not necessary. But since it is an `OPTIONAL` clause it does not break anything.

wolfderechter:
> Looks good. Can you add the tickets to the next release on Jira as well?

Done 

> **Comment (non-blocking):** Put unrelated changes into separate PRs.

I completely agree, I only bundled them now because of the high prio of the tickets, and because of the limited changed which would have been annoying to setup 3 times :p

PR #549, opened 2025-06-03
fix: make relations between provinces and municipalities readable for all users

mirdono:
The frontend relies on the memberships relating municipalities to their provinces to
- automatically fill in the correct province in forms when a municipality was selected; and
- automatically limit the allowed municipalities when a province is selected in a form.

In #541 the necessary membership data for this functionality was moved to the `administrative-unit` graph, making it inaccessible to users with a worship role. Consequently, the functionality is broken for such users. This prevents them, for example, from creating new worship services as they cannot fill in the complete address.

## Proposed solution
Move the membership relating provinces and municipalities back to the `shared` graph so that users of the worship module are unblocked.

Note that this does reintroduce the issues described #541, with memberships that cannot be removed and possibly duplicate memberships being added in different graphs. I propose to raise these latter issues with business and look into a better solution for them.

## How to reproduce the reported bug
The bugs only occur when logged in with a **worship** role. To compare with the expected behaviour you can log in with a non-worship role.

### Address forms
1. Log in to OP with worship editor role
2. Use the "+ Nieuw" button in the top-right corner to go to the form for creating a new worship organisation
3. In the "Primaire contactgegevens" section, enter an address (any will do)
4. Under normal operation the "Provincie" field should be automatically filled in with the corresponding value.

Alternatively, manually enter an address
3. In the "Primaire contactgegevens" section click "Vul adres manueel in"
4. Select a province in the bottom field, any will do
5. Try to select a municipality, no entered search will give any results. Under normal operation the provided municipality options should be limited to those in the selected province.
6. Clear the selected province, and select a municipality in the appropriate field, any will do
7. Under normal operation the "Provincie" field should be automatically filled in with the corresponding value.

Alternatively, try to create a new site for an existing organisation. This uses the same form which therefor has the same issue.

### Sidebar on organizations page
1. Log in to OP with a worship role
2. In the sidebar select a municipality, any will do, in the corresponding filter.
3. Under normal circumstances the province filter is automatically filled in with the correct province. Now it remains empty and it does not provide any options to select.
4. Clear the selected municipality.
5. Select a province in the corresponding filter, any province will do
6. Under normal operation the options in the municipality filter should be limited to those located in the selected province. Now it will never show any options.

## Related tickets
- OP-3619

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Deploy notes
 ```
 drc restart migrations; drc logs -ft --tail=200 migrations
+drc exec db-cleanup curl -X POST "http://localhost/cleanup"
 ```
 
 ## v1.32.1 (2025-06-02)
--- a/config/migrations/2025/20250603161306-fix--move-memberships-for-municipalities-and-provinces-to-shared/20250603163337-feat--cleanup-job-insert-suborganization-province-municipality.graph
+++ b/config/migrations/2025/20250603161306-fix--move-memberships-for-municipalities-and-provinces-to-shared/20250603163337-feat--cleanup-job-insert-suborganization-province-municipality.graph
@@ -1,0 +1,1 @@
+http://mu.semte.ch/graphs/public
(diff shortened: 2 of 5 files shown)


claire-lovisa (approved):
I also had to `drc restart cache resource` after the db cleanup, but it works well. It's a good idea to keep track of potential solutions for the issue that we introduce back here, but that fix will do just fine in the meantime :)

PR #558, opened 2025-07-02
fix: organisations report

mirdono:
The generated organisation reports contained entries that were missing a value in the label column, a mandatory field for organisations. The underlying issue was that the queries specified two resource types: `org:Organization` and `besluit:Bestuurseenheid`. This caused multiple entries in the query results for organisations that have their type duplicated in both the `shared` and `administrative-unit` graph, where one entry was lacking a value for the label field. When merging the query results into a report file only one entry remained, typically the one lacking the necessary data.

## Proposed solution
Remove the `besluit:Bestuurseenheid` as type to query for. This is not needed as all administrative units should also have the `org:Organization` type. This also makes to use of `VALUES` clauses unnecessary as `?type` will only be set to a single value.

In the process of verifying the resulting reports I noticed that in some cases organisations were missed by the query, for DEV this was about 50 organisations. Switching the order of the `GRAPH` and `OPTIONAL` clauses as well as moving the corresponding graph `FILTER` into the `OPTIONAL` clause resolved that without losing data (tested with DEV and PROD data).

## How to test
1. Launch to OP stack
2. Bump the `report-generation-service` to the new version
3. Generate a report, for example by executing:

``` shell
docker compose exec report-generation curl -X POST localhost/reports -d "{\"data\": {\"attributes\": {\"reportName\": \"organizations\"}}}" --header "Content-Type: application/json"
```

4. A new report file should have been created in `/data/files/exports/`. Inspect that file to see whether it has the right amount of entries and no entries without a value in the label column.

To see how organisation resources are in your data, thus how many there there should be in the report:

``` sparql
PREFIX org: <http://www.w3.org/ns/org#>

SELECT count(DISTINCT ?s)
WHERE {
  GRAPH ?g {
    ?s a org:Organization .
  }
  VALUES ?g {
    <http://mu.semte.ch/graphs/worship-service>
    <http://mu.semte.ch/graphs/shared>
  }
}
```

## Open questions
- Do the duplicated type triples in the two graphs serve a purpose? If not, we might consider removing the from the `administrative-unit` graph altogether.
- The `skos:prefLabel` predicate is mandatory for organisations, and all organisation resources in DEV and PROD have it. What is the reason to make it optional in the query?

## Other changes
- Bumped the report-generation service to the latest version (first commit)
- Formatted the organisations report generation configuration (second commit)

## Related tickets
- OP-3636

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,14 @@
 - Add Brussel as a municipality [OP-3628]
 - Bump delta-notifier service to most recent version [OP-3190]
 - Add URI to resource's organization responses
+- Fix query for organizations report [OP-3636]
+- Bump report-generation-service to most recent version
 ### Deploy notes
 ```
 drc pull deltanotifier; drc up -d deltanotifier
 drc restart migrations-triggering-indexing resource
 drc pull frontend; drc up -d frontend
+drc pull report-generation; drc up -d report-generation
 ```
 
 ## v1.33.2 (2025-06-20)


mirdono:
> I tested and it works fine, I just have a question, do you know where the triple-quote escape comes from? It it because of the bump of the report service? I'm a bit scared it would confuse the people reading those reports, maybe there is an option to switch back to simple-quote escapes?

Not sure but there were indeed some changes related to quotes and escaping ([#11](https://github.com/lblod/loket-report-generation-service/pull/11) and this [commit](https://github.com/lblod/loket-report-generation-service/commit/a1ffd7cd4ce33ad728d9f193a6c2d37beb3ac0ba)) in the service. As far as I can see there is no option to change or configure it to use simple-quote escapes again. I am not sure who uses these reports, I was assuming on Sofie does.

> 
> Now about your open questions:
> > Do the duplicated type triples in the two graphs serve a purpose? If not, we might consider removing the from the administrative-unit graph altogether.
> 
> For this I don't really know, we could try to see how the app behaves

Okay, that sounds like a separate ticket than.

> 
> > The skos:prefLabel predicate is mandatory for organisations, and all organisation resources in DEV and PROD have it. What is the reason to make it optional in the query?
> 
> I think it's as a measure of safety, in case it would disapear for some reason and went unnoticed, the report would still show the associated line

That sounds reasonable. I will refrain myself from removing the `OPTIONAL` :wink:

PR #559, opened 2025-07-03
Add migrations to revert Brussels' province and municipality

claire-lovisa:
# Context

[OP-3637]

In this PR, we revert the changes made in #553, #555 and #557 because it has been decided after some discussion that the model we started to use is not the one we should be continuing with. The new model will come later on, for now we just revert to avoid propagating changes we don't want to other environments.

**Note:** the frontend update to prevent the user to look for an address in Brussels via the address search component is kept, this was causing validation issues regardless of those changes.

# Test instructions

Let run the `migrations-triggering-indexing` service, give the `search` service a few minutes to update the indexes as well. You should not be able to find the province or municipality of Brussel, the addresses should be back to belonging to "Vlaams-Brabant" and all the added data should be gone.

### DIFF
--- a/config/resources/domain.json
+++ b/config/resources/domain.json
@@ -389,10 +389,6 @@
         "purpose": {
           "type": "string",
           "predicate": "org:purpose"
-        },
-        "note": {
-          "type": "string",
-          "predicate": "skos:note"
         }
       },
       "relationships": {


mirdono (commented):
**Question:** Impact on deltas from insert and immediate deletion?

On QA and PROD we will now insert data and, almost, immediately remove it again. Do we risk any weird delta behaviour with this?

mirdono:
> @mirdono Thanks for taking the time to review :) About the impact of insert/delete for QA and PROD, what will most likely happen is that it'll create delta files to produce it and immediately after to remove it, causing consuming apps to consume and delete the data. Now that you point it out, it's probably suboptimal. But we also need to run it on dev. Could we envision a deploy to dev, immediately followed by another PR removing those migrations completely? So they never even have to run on QA/PROD, but we still have a trace via PRs of what happened?

I see two options:
- Add a third PR after this one that removes all the migrations added by #553, #555, #557, and the current PR. It does require that all these PRs are part of the same release, but since there is no release going on atm that should not be an issue.
- Run the migrations in this PR as local migrations on DEV and convert this PR to one removing the migrations from #553, #555,and  #557. This requires a PR less, but could be considered as having lower traceability. One would have to look at the unversionned migrations on DEV, and associate them to these PRs, to fully understand the data was added and removed again. (A remark in this PR pointing to the use of local migrations could improve that situation.)

mirdono (approved):

PR #564, opened 2025-07-14
Add migration to fix dev addresses

claire-lovisa:
# ID

[DL-6701]

# Description

On dev, some addresses are linked both to sites and to contact point of mandatarissen. It doesn't make any sense and I couldn't trace it back to the source. The issue did not affect other environments.

The migration should fix it by creating a dupe of the affected addresses and by linking those dupes to the sites. That way, sites and contact points have independent addresses. We update the link between site and address and not between mandataris and address because OP is master of the sites, which means we should then produce the correct data and forward it to consuming apps.

### DIFF
--- a/config/dev-migrations/20250708142100-fix-same-address-linked-to-both-site-and-contact-point.sparql
+++ b/config/dev-migrations/20250708142100-fix-same-address-linked-to-both-site-and-contact-point.sparql
@@ -1,0 +1,31 @@
+DELETE {
+  GRAPH ?h {
+    ?site <https://data.vlaanderen.be/ns/organisatie#bestaatUit> ?oldAddressUri .
+  }
+} INSERT {
+  GRAPH ?g {
+    ?newAddressUri a <http://www.w3.org/ns/locn#Address> ;
+      <http://mu.semte.ch/vocabularies/core/uuid> ?newAddressUuid ;
+      ?p ?o .
+  }
+
+  GRAPH ?h {
+    ?site <https://data.vlaanderen.be/ns/organisatie#bestaatUit> ?newAddressUri .
+  }
+} WHERE {
+  GRAPH ?g {
+    ?oldAddressUri a <http://www.w3.org/ns/locn#Address> ;
+      ?p ?o .
+
+    FILTER (?p NOT IN (<http://mu.semte.ch/vocabularies/core/uuid>))
+  }
+
+  ?mandatarisContact <http://www.w3.org/ns/locn#address> ?oldAddressUri .
+
+  GRAPH ?h {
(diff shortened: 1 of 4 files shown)


config/dev-migrations/20250708142100-fix-same-address-linked-to-both-site-and-contact-point.sparql
```diff
@@ -0,0 +1,31 @@
...
+    ?oldAddressUri a <http://www.w3.org/ns/locn#Address> ;
+      ?p ?o .
+
+    FILTER (?p NOT IN (<http://mu.semte.ch/vocabularies/core/uuid>))
+  }
+
+  ?mandatarisContact <http://www.w3.org/ns/locn#address> ?oldAddressUri .
+
+  GRAPH ?h {
+    ?site <https://data.vlaanderen.be/ns/organisatie#bestaatUit> ?oldAddressUri .
```
cecemel, line 26: nitpick: maybe just to have a precise query, add a `a <http://www.w3.org/ns/org#Site>`

config/dev-migrations/20250708142100-fix-same-address-linked-to-both-site-and-contact-point.sparql
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  BIND(MD5(CONCAT(?oldAddressUri,"OLD_ADDRESS")) AS ?newAddressUuid) .
```
cecemel, line 29: nitpick, cause i was a bit confused, add `"NEW_ADDRESS"` as seed :sweat_smile:

config/dev-migrations/20250708142100-fix-same-address-linked-to-both-site-and-contact-point.sparql
```diff
@@ -0,0 +1,31 @@
...
+    ?site <https://data.vlaanderen.be/ns/organisatie#bestaatUit> ?oldAddressUri .
+  }
+} INSERT {
+  GRAPH ?g {
+    ?newAddressUri a <http://www.w3.org/ns/locn#Address> ;
+      <http://mu.semte.ch/vocabularies/core/uuid> ?newAddressUuid ;
+      ?p ?o .
+  }
+
+  GRAPH ?h {
```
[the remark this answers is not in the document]
claire-lovisa: ooh that's a good catch, pretty strange data. We don't have it in prod, but I now see that we have it on QA, same for the addresses shared by a site and contact points. Not entirely sure how I missed it last time I checked, I looked at the recent updates and see nothing that should impact this, so it's probably my mistake. I'll move the migration to the regular folder and add a migration to move addresses that are in the wrong graph.

PR #566, opened 2025-07-25
add duplicate memberships cleanup migration

wolfderechter:
## ID

OP-3634

## Description

Following the conclusion of the analysis (see ticket), the solution of the duplicate memberships should be to remove the inverse membership where the municipality is linked with `org:member` and the ocmw is linked with `org:organization`.

## Testing

Using op dev data, the query should remove all the inverse/duplicate memberships. For example of 'gemeente affligem' it should only have 1 membership with 'ocmw affligem'. The same membership should show up for 'ocmw affligem' as well.

### DIFF
--- a/config/migrations/2025/20250725120522-remove-duplicate-memberships-ocmw-municipality.sparql
+++ b/config/migrations/2025/20250725120522-remove-duplicate-memberships-ocmw-municipality.sparql
@@ -20,14 +20,14 @@ DELETE {
       org:role <http://data.lblod.info/id/rollen/4ec7d5c39bdc4e84b4174379b9e22ad8> .
   }
 
-  GRAPH ?graphMunicipality {
+  GRAPH <http://mu.semte.ch/graphs/shared> {
     ?municipality org:classification ?municipalityClassification .
   }
 
-  GRAPH ?graphOCMW {
-    ?ocmw org:classification ?ocmwClassificiation .
+  GRAPH <http://mu.semte.ch/graphs/administrative-unit> {
+    ?ocmw org:classification ?ocmwClassification .
   }
 
   FILTER(?municipalityClassification = <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000001>)
-  FILTER(?ocmwClassificiation = <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000002>)
+  FILTER(?ocmwClassification = <http://data.vlaanderen.be/id/concept/BestuurseenheidClassificatieCode/5ab0e9b8a3b2ca7c5e000002>)
 }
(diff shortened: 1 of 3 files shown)


claire-lovisa (changes requested):
I'm getting a very strange behaviour :sweat_smile: I don't see what the issue is at first sight but after the migration **and restarting the whole stack** (if I just restart migrations, it's fine), I can't open any admin unit anymore. I'm being redirected to http://localhost:4200/pagina-niet-gevonden . Maybe the frontend expects something that this migration removes? Can you also reproduce that behaviour?

wolfderechter:
@claire-lovisa I've found the culprit, it was the frontend that was not released yet. If you use the latest tag it should work fine

mirdono:
> @claire-lovisa I've found the culprit, it was the frontend that was not released yet. If you use the latest tag it should work fine

For future reference. Due to [model changes](https://github.com/lblod/app-organization-portal/pull/550/files#diff-cfe6fd77de75828314f7d51475e16bfce1a8eebd7e7cf5dd266609559fc05254) as part of the new edit scope of operation functionality, the tagged frontend versions (v1.33.0 and earlier) are not compatible at the moment with with the development branch of the backend. This is causing the observed errors. This can be avoided by using the latest tag, as already mentioned by @wolfderechter, or locally building the frontend from its development branch.

To avoid such confusion in the future, we could consider to stop using the `latest` tag for the frontend on DEV and use tagged versions instead. In other words threat the frontend the same way as we do other services. (I think this is also more in line with what other apps do.) But this probably requires a broader discussion.

config/migrations/2025/20250725120522-remove-duplicate-memberships-ocmw-municipality.sparql
```diff
@@ -0,0 +1,33 @@
...
+  }
+} WHERE {
+  GRAPH <http://mu.semte.ch/graphs/administrative-unit> {
+    ?duplicateMembership a org:Membership ;
+      org:organization ?orgB ; # OCMW
+      org:member ?orgA ; # Municipality
+      org:role <http://data.lblod.info/id/rollen/4ec7d5c39bdc4e84b4174379b9e22ad8> ;
+      ?p ?o .
+
+    # Find the other memberships with swapped organizations
```
mirdono, line 16: **Comment:** Requiring a match for the "correct" membership can cause "duplicate" memberships to be missed.

For example, on DEV Aalst municipality and OCMW are only related via a "duplicate" membership, the other one probably got removed while testing something. After running this migrations that membership remains in place. If users did something similar on PROD we could run into situations where we end up with "duplicate" memberships instead of the correct ones.

config/migrations/2025/20250725120522-remove-duplicate-memberships-ocmw-municipality.sparql
```diff
@@ -0,0 +1,33 @@
...
+PREFIX mu: <http://mu.semte.ch/vocabularies/core/>
+
+DELETE {
+  GRAPH <http://mu.semte.ch/graphs/administrative-unit> {
+    ?duplicateMembership ?p ?o .
+  }
+} WHERE {
+  GRAPH <http://mu.semte.ch/graphs/administrative-unit> {
+    ?duplicateMembership a org:Membership ;
+      org:organization ?orgB ; # OCMW
```
mirdono, line 11: **Nitpick:** Using more meaningful variable names makes the query easier to understand.

Imo using more meaningful names such as `?municipality[Org]` instead a generic name such as `?orgA` makes the query easier to understand. Now one has to mentally map `A` and `B` to `municipality` and `ocmw` respectively while reading the query.

PR #568, opened 2025-08-04
Import municipalities of Brussel and fix province value of related addresses

claire-lovisa:
# Context

[OP-3638] & [OP-3648]

The decision taken to handle provinces for addresses outside of Flanders states that:
- we shouldn't link a province to addresses outside of Flanders
- we should be able to manually enter addresses outside of Flanders (and even outside of Belgium)
- a province is still mandatory for addresses within Flanders

This PR aims to:
1. get rid of the province of addresses outside of Flanders (after data analysis, only data within the Brussel's area is concerned, this is why the migration focuses on those)
2. add data about municipalities of the Brussel's area in OP, reusing URIs created by Wegwijs (because they should be out of the Flemish's "jurisdiction")

# How to test?

First, start up the stack with the `migrations-triggering-indexing` service disabled:
```
  migrations-triggering-indexing:
    entrypoint: ["echo", "Service disabled"]
```

Once the stack has fully started, enable it again and `drc up -d`. Wait for the `search` service to completely index newly created data.

You should now be able to see the new municipalities via the search engine as well as view/edit their data, and in the main tables you should see an empty "Provincie" column when the city is Brussel.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - datafix: cleanup of general memberships if specific ones exist [OP-3640]
 - datafix: cleanup of duplicate memberships [OP-3634]
 - Fix KBO organizations' date formats [OP-3560]
+- Add municipalities in the area of Brussel and remove province for addresses in this area [OP-3638] & [OP-3648]
```
mirdono, line 13: **Issue:** Add command to restart `migrations-triggering-indexing` to deploy notes.

The deploy notes currently lack a command to restart this service such that the added migrations are executed.
Side note, due a release in the meantime the changelog has changed on development so you might update this branch with respect to `development`.

config/migrations-triggering-indexing/2025/20250728171200-brussels-municipalities/20250731153700-insert-anderlecht-municipality.sparql
```diff
@@ -0,0 +1,77 @@
+INSERT DATA {
```
mirdono, line 1: **Question:** Should we also create a placeholder for the SharePoint identifier.

When creating an organisation via the frontend there is also some boilerplate triples inserted for the Sharepoint identifier, irrelevant of whether a value was entered by the user. More specifically, the `adms:Identifier` and `generiek:GestructureerdeIdentificator` resources are already created. When adding new organisation via migrations I always did the same.
Is this not necessary since you do not do it here? Based on some quick testing it does not cause any errors and the appropriate triples are created when filling the SP id via the frontend.

mirdono (changes requested):
Encountered one small issue and two questions.

**Question:** Should the new municipalities show in the organisations overview table and should they be editable?

Currently the newly added municipalities end up being included in the organisations overview along with all other organisations. (I did have to do a full re-index, merely executing the migrations and waiting "long" enough did not seem to do it.) Is this the intended behaviour. Furthermore, should these new municipalities be editable in the same as other municipalities? As you mention the new ones do not fall under Flemish jurisdiction. From an OP standpoint it imo seems a bit weird to threat them the same as the Flemish municipalities (which includes producing this data again to others.)

config/migrations-triggering-indexing/2025/20250728171200-brussels-municipalities/20250731153700-insert-anderlecht-municipality.sparql
claire-lovisa: It's not necessary I think, the frontend creates it automatically when it's missing https://github.com/lblod/frontend-organization-portal/blob/development/app/routes/organizations/organization/core-data/edit.js#L68 , which explains why you didn't have any error when filling it in

claire-lovisa:
@mirdono Your questions were on point and triggered a whole discussion. For now, we don't want to show those municipalities on the organizations page, as it could confuse end users. The solution that we decided on for now is to use the same behavior as municipalities of foreign countries: instead of selecting the municipality from the list of all the municipalities we have in our organizations, we simply enter it as a string. This also has the benefit of allowing users to enter addresses in Wallonia, if they would need it at some point. The code of this PR as well as https://github.com/lblod/frontend-organization-portal/pull/688 has been updated accordingly.

mirdono (approved):

PR #591, opened 2026-04-09
Add EredienstBeroepen and TypeBetrokkenheid labels

bdevloed:
OP-3771 Adds labels for:

- EredienstBeroepen
- TypeBetrokkenheid

### DIFF
--- a/config/delta-producer/public/export.json
+++ b/config/delta-producer/public/export.json
@@ -526,6 +526,32 @@
       ]
     },
     {
+      "type": "http://lblod.data.gift/vocabularies/organisatie/EredienstBeroepen",
+      "graphsFilter": [
+        "http://mu.semte.ch/graphs/public"
+      ],
+      "pathToConceptScheme": [
+        "http://www.w3.org/2004/02/skos/core#inScheme"
+      ],
+      "properties": [
+        "http://mu.semte.ch/vocabularies/core/uuid",
+        "http://www.w3.org/2004/02/skos/core#prefLabel"
+      ]
+    },
+    {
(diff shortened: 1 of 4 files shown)


bdevloed:
Adding copilot as reviewer was done by accident.

config/delta-producer/public/export.json
```diff
@@ -525,6 +525,32 @@
         "http://www.w3.org/2004/02/skos/core#prefLabel"
       ]
     },
+    {
+      "type": "http://lblod.data.gift/vocabularies/organisatie/EredienstBeroepen",
+      "graphsFilter": [
+        "http://mu.semte.ch/graphs/public"
+      ],
+      "pathToConceptScheme": [
+        "http://www.w3.org/2004/02/skos/core#inScheme"
```
DamonKennes, line 534: I think the link to the public export conceptScheme is not in the data for these two types.
 
"?p <http://www.w3.org/2004/02/skos/core#inScheme> <http://lblod.data.gift/concept-schemes/10caf4a4-1ff5-4c05-917e-56489910d68b>" 

Because of this EredienstBeroepen and TypeBetrokkenheid seems to not get picked up. Maybe we can add a similar migration like 20250306113500-add-concept-scheme-to-codes-for-public-producer.sparql for these two types.

DamonKennes (changes requested):

config/delta-producer/public/export.json
cecemel: @DamonKennes good catch. I added it. Maybe a feature to check if combination can work too (i.e. with and without path to conceptscheme)

cecemel (approved):

PR #599, opened 2026-04-23
DL-6575: added sparql-parser + edit changelog

DamonKennes:
- changed mu-auth to sparql-parser


NikiDeVisscher (commented):
Looks good, seems to work.
A few apps that received the sparql-parser update had some problems in production because of outdated mu-cl-resources and sparql parser being faster and generating a lot of output. Best to update and add these as well here, see https://github.com/lblod/app-lpdc-digitaal-loket/pull/96#pullrequestreview-4081153316

NikiDeVisscher (approved):

PR #603, opened 2026-06-17
Added juridische vorm + change-event

DamonKennes:
- Added Juridische vorm [OP-3816]
- Added change-event to change juridische vorm [OP-3820]

NOTE:
- Made juridische vorm not required because migration to fill in existing organizations will come later
- Juridische vorm is needed for organizations and registered organizations except worship administrative units. At the moment i put it on organizations because I don't really see a way of adding it elsewhere. This seems a bit weird because worship administrative units is a subclass of organizations.
- Regarding the change-event, I looked at the existing code regarding the status change via change-events and implemented it in a similar way. I think maybe it would be interesting to look at a change-event-service that would make changes via a config based on a newly created change-event delta.

### DIFF
--- a/config/migrations/2026/20260615120000-add-legal-forms-codelist.ttl
+++ b/config/migrations/2026/20260615120000-add-legal-forms-codelist.ttl
@@ -1,0 +1,188 @@
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix mu: <http://mu.semte.ch/vocabularies/core/> .
+@prefix ext: <http://mu.semte.ch/vocabularies/ext/> .
+@prefix conceptscheme: <http://data.vlaanderen.be/id/conceptscheme/> .
+@prefix concept: <http://data.vlaanderen.be/id/concept/LegalFormCode/> .
+
+# Concept Scheme for Legal Forms
+conceptscheme:LegalFormCode
+  rdf:type skos:ConceptScheme ;
+  skos:prefLabel "Juridische vorm" .
+
+# Legal Form Codes
+concept:4a55b679-e4c7-4014-a0b8-3d6762e0225f
+  rdf:type ext:LegalForm ;
+  rdf:type skos:Concept ;
+  mu:uuid "4a55b679-e4c7-4014-a0b8-3d6762e0225f" ;
+  skos:prefLabel "Stad / gemeente" ;
+  skos:topConceptOf conceptscheme:LegalFormCode ;
+  skos:inScheme conceptscheme:LegalFormCode .
+
+concept:e87b0707-1f9c-4577-8918-023e60974d2f
+  rdf:type ext:LegalForm ;
+  rdf:type skos:Concept ;
+  mu:uuid "e87b0707-1f9c-4577-8918-023e60974d2f" ;
+  skos:prefLabel "Openbaar centrum voor maatschappelijk welzijn" ;
+  skos:topConceptOf conceptscheme:LegalFormCode ;
+  skos:inScheme conceptscheme:LegalFormCode .
+
+concept:fa7b4356-6e67-4d3a-9256-72922a12227a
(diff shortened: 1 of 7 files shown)


bdevloed (changes requested):
`ext:LegalForm` -> `code:Rechtsvormtype`

config/resources/domain.json
```diff
@@ -942,6 +947,11 @@
       },
       "new-resource-base": "http://lblod.data.gift/concepts/"
     },
+    "legal-form-codes": {
```
Windvis: Is there a specific reason why we need a new resource instead of just using `concept` directly? It would simplify things, no? (no dispatcher config, no extra resource config, no extra frontend models, ..)

DamonKennes: I also considered whether this could be implemented using the concepts resource directly instead of creating a new resource. I think it should be technically possible to work with just concepts.

I decided to go for a new resource because I thought the data model would not really reflect reality. The organisatie:rechtsvorm triple is only valid for a specific type of concept (the legal form ones) and having a dedicated resource type makes this constraint explicit. I also saw that other similar resources like board-position-codes and organization-classification-codes are implemented in a similar way.

So I think it's possible to implement this with just the generic concept resource but I'm not sure what would be best. @bdevloed do you have a preference regarding this?

config/migrations/2026/20260615120000-add-legal-forms-codelist.ttl
```diff
@@ -0,0 +1,212 @@
...
+@prefix conceptscheme: <http://data.vlaanderen.be/id/conceptscheme/> .
+@prefix concept: <http://data.vlaanderen.be/id/concept/LegalFormCode/> .
+
+# Concept Scheme for Legal Forms
+conceptscheme:LegalFormCode
+  rdf:type skos:ConceptScheme ;
+  skos:prefLabel "Juridische vorm" .
+
+# Legal Form Codes
+concept:4a55b679-e4c7-4014-a0b8-3d6762e0225f
```
Windvis, line 16: I wonder if we should add "order" predicates to these so we can sort on in from the frontend. The "andere" option is just positioned randomly while most users would expect that at the end I think?

Alternatively we can maybe just sort on label or something, so the order doesn't depend on the way Virtuoso gives it back to us without providing a sort attribute.

DamonKennes: I added ordering via the migration based on the order in the ticket. "Andere" not being at the bottom seems indeed not very user friendly so I kept that one last.

config/resources/domain.json
Windvis: FYI, this isn't a blocker for me. I just know there was a historical reason for doing it this way (something mu-cl-resource related) and we started to use "concept" directly in newer code. But you have a point that it's more explicit which values are allowed with the custom type. 

I don't think it's worth the extra overhead of having to add resource config, dispatcher config, frontend code, etc, but it's not _wrong_ either. So since the code is already there it might be better to keep it as-is and get this merged? Or where there other things you still wanted to change?

DamonKennes: I checked with @bdevloed and he also thinks changing it to concept would be better. He first thought this seperate legal form resource entity already existed in the codebase. After looking around I'm not able to found it anymore although it seems something similar existed at one time in the code history.

I'm currently looking at changing the code to use the concept resource directly. I'll update the PR with the change afther that it should be ready to merge.

config/resources/domain.json
```diff
@@ -938,10 +943,32 @@
         "alt-label": {
           "type": "string",
           "predicate": "skos:altLabel"
+        },
+        "order": {
+          "type": "number",
+          "predicate": "ext:order"
```
Windvis: I think we normally use a different predicate for "orders" but forgot which one 🤔.

DamonKennes: I updated it to use sh:order. Found this one being used in other applications like [app-lokaal-mandatenbeheer](https://github.com/lblod/app-lokaal-mandatenbeheer)

